PR #9, opened 2023-10-06
Adding Spy and Investigating Leads

LezaiYa:


Roles/Crewmate/Spy.cs
```diff
@@ -0,0 +1,100 @@
...
+        var NormalColor = taskState.IsTaskFinished ? TaskCompleteColor : NonCompleteColor;
+        TextColor = comms ? Color.gray : NormalColor;
+        string Completed = comms ? "?" : $"{taskState.CompletedTasksCount}";
+
+        Color TextColor1;
+        if (UseLimit[playerId] < 1) TextColor1 = Color.red;
+        else TextColor1 = Color.white;
+
+        sb.Append(ColorString(TextColor, $"<color=#777777>-</color> {Completed}/{taskState.AllTasksCount}"));
+        sb.Append(ColorString(TextColor1, $" <color=#777777>-</color> {Math.Round(UseLimit[playerId], 1)}"));
```
Tommy-XL, line 95: It looks like tasks and abilities are showing up as "-0/5- 2" when they should be "(0/5) - 2"

LezaiYa: Now it's done.

Tommy-XL:
Everything else seems fine

PR #14, opened 2023-10-06
Updated mini, custom eviction text and Chinese announcements

LezaiYa:


Patches/PlayerControlPatch.cs
```diff
@@ -2698,14 +2700,16 @@ public static void Postfix(PlayerControl __instance)
...
 
                             if (Mini.UpDateAge.GetBool())
                             {
-                                Mini.SendRPC();
-                                Utils.NotifyRoles();
+                                foreach (var pc in Main.AllPlayerControls)
+                                {
+                                    if (pc.PlayerId != player.PlayerId) continue;
+                                    player.Notify(GetString("MiniUp"));
+                                }
```
Tommy-XL, line 2710: I don't think it's necessary to run foreach

You can just write:
`if (player.Is(CustomRoles.Nice/EvilMini)) 
player.Notify(GetString("MiniUp"));`

Tommy-XL (approved):

Patches/PlayerControlPatch.cs
LezaiYa: It's done. _.

PR #38, opened 2023-10-08
Updated shaman, reworked Reverie, removed glow and some other changes

ryuk201198:
1. updated shamn text string (removed coven info)
2. Removed gradient tags (cause of DC issues) and updated strings accordingly
3. Changed Modded NA url


ryuk201198:
removed glow, reworked reverie and some fix in the latest commit
1. removed glow
2. Reworked reverie
3. president showing /reveal fixed
4. removed /cp as command for inspector (ask moe why)

PR #62, opened 2023-10-13
Stop Kicking Out Developers

LezaiYa:


ryuk201198:
Ppl can spoof friend code of devs and simply become unbannable.

PR #84, opened 2023-10-15
Instigator

lars-wu:
InstigatorInfoLong:
As the Instigator, it's your job to turn the crewmates against each other. Each time a Crewmate is voted out in a meeting, as long as you are alive, an additional Crewmate who voted for the innocent player will die after the meeting. The number of additional players dying is determined by the host


0xDrMoe:
This has been converted to Draft as we are not doing role pushes until next full release

Patches/ExilePatch.cs
```diff
@@ -1,11 +1,10 @@
+using System.Linq;
 using AmongUs.Data;
 using HarmonyLib;
-using System.Linq;
```
0xDrMoe, line 3: why did this get removed

Resources/Lang/en_US.json
```diff
@@ -540,6 +541,7 @@
   "BerserkerInfo": "Kill to increase your level",
   "EvilMiniInfo": "No one can hurt you until you grow up",
   "BlackmailerInfo": "Silence other players",
+  "InstigatorInfo": "Sow discord among the crewmates",
```
0xDrMoe, line 544: Sow? I'm so lost loool

0xDrMoe (changes requested):

Patches/ExilePatch.cs
Tommy-XL: Maybe it's just not used in the code

Gurge44: When you do an auto code cleanup in VS, it removes unnecessary using directives

lars-wu: What @Gurge44 said.

Resources/Lang/en_US.json
lars-wu: Is it not correct english? haha. How would you say it? Translator says it's fine.

Ultradragon005: sow is correct, but simply do "Cause discord among the crewmates" imo

Patches/ExilePatch.cs
```diff
@@ -214,7 +217,8 @@ CustomRoles.Tracefinder or
                 CustomRoles.Doctor or
                 CustomRoles.Alchemist or
                 CustomRoles.Bomber or
-                CustomRoles.Undertaker
+                CustomRoles.Undertaker or
+                CustomRoles.Instigator
                 ) pc.RpcResetAbilityCooldown();
```
Tommy-XL, line 223: I'll temporarily remove this code because I've re coded it to #206

Tommy-XL (approved):

Tommy-XL:
Need fix conflict

PR #86, opened 2023-10-16
Try fix admirer

NikoCat233:
1.fixed the bug where admirer uses a shared limit
2.fixed neutral wins in exile patch
3.Add checks in set role process (imagine setting a neutral to be recruit / madmate)
4.Now we can set the skill limit of admirer

Neutral killers wins arent fixed yet(This gonna rework check game end)

### DIFF
--- a/Modules/Utils.cs
+++ b/Modules/Utils.cs
@@ -982,7 +982,7 @@ public static class Utils
                     ProgressText.Append(CursedSoul.GetCurseLimit());
                     break;
                 case CustomRoles.Admirer:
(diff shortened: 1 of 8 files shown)


NikoCat233:
Tommy can u help me add null check for utils.getplayerbbyid?
Sometimes the exiled player quits and left a null pc, bugging the game.

Gurge44:
Literally just add a question mark after "pc" where a method is used with it

NikoCat233:
> Literally just add a question mark after "pc" where a method is used with it

Thanks

PR #89, opened 2023-10-16
Hater no longer ignores shields & Code improvement

Gurge44:
I don't have time to test anything, so I would apprechiate if someone could. Thanks.


NikoCat233:
Im testing it later

NikoCat233:
Bug:
1.shield wont take off after FFF's kill attempt
2.FFF still wins even it target got shield protects

Bruh idk how 2 fix this

Gurge44:
Well yea.... the whole winner condition code needs to be moved into MurderPlayerPatch. It would make it a lot easier to work with and we could avoid using RpcCheckAndMurder(), since the code will only run when the Hater actually kills. I'll make this when I get home.

NikoCat233:
Suggest directly move check murder 2 murder player

NikoCat233:
Also the string of FFF needs to be updated

NikoCat233:
No not me , I got classes to attend the whole day =(

NikoCat233:
Tested and it works fine! Just add more logger

And thanks for the detailed comments, You truly made me understand everything

Gurge44:
I think it can be merged if others are ok with it, you can add more logging later

PR #91, opened 2023-10-17
Recode Imitator, Amnesiac. Fix admirer in copycat

NikoCat233:
1.Recode Imitator
2.Remove useless codes in Amnesiac
3.Fix add admirer in copycat
4.Fix some KnowRole bug in know role target(This allows some role to see its companions which is not supposed)

Who f* copy paste from Imitator to Amnesiac? I litterly coded on Amnesiac check murder and realized the fact that Amnesiac dont use check murder when i finished. Just to say , F* those cv WORKER without checking codes. 
And who f* coded these trash codes just by cv 100 times?

### DIFF
--- a/Roles/Crewmate/Admirer.cs
+++ b/Roles/Crewmate/Admirer.cs
@@ -18,7 +18,7 @@ public static class Admirer
     public static OptionItem AdmireCooldown;
     public static OptionItem KnowTargetRole;
     public static OptionItem SkillLimit;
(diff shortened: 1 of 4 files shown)


ryuk201198:
Lol blame loonie.

NikoCat233:
Even if we are not using Imitator, this can still be merged bcz it fixs Amnesiac and some other bugs

Tommy-XL (approved):
Seems, everything to be fine

PR #99, opened 2023-10-19
added sponsor, temporary delayed spam fix, fix wrong name shown for bait reveal and /id, and admirer fix, and spy fix

ryuk201198:
1. Added sponsor
2. Temporary fix for delayed spamming [INCLUDES THE IMPROVED SPAMMING SYSTEM] (first spams using OG system and then spams using new system)
3. fix /id, /mid and bait reveal showed the original name of the player when using color name mode
4. fixed spy not showing orange names for modded clients.


ryuk201198:
fixed spy not showing orange name for modded client
KNOWN BUG : - once name turns orange it doesnt change back because removing RPC is not received by the modded client.

ryuk201198:
Fixed spy. I didnt close the Sent RPC like an idiot that is why modded client wasnt receiving the rpc 🥲

Tommy-XL (approved):

PR #101, opened 2023-10-20
Merchant setting

ryuk201198:
Setting to sell only enabled addons


Roles/Crewmate/Merchant.cs
```diff
@@ -132,10 +137,16 @@ public static void OnTaskFinished(PlayerControl player)
             {
                 return;
             }
+            if (addons.Count <= 0)
```
Tommy-XL, line 140: Here you can use `!addons.Any()` this is more performant than `.Count <= 0`

ryuk201198: I see

ryuk201198:
@Tommy-XL i changed count to any and did it a couple of other places too.

Tommy-XL (approved):

PR #103, opened 2023-10-20
Code Improvements (Part 1)

Tommy-XL:
Don't merge this until I'm done, mostly I need to see the changes I've made to the code and also someone can suggest ideas for improvement


Patches/CheckGameEndPatch.cs
```diff
@@ -491,7 +491,7 @@ public bool CheckGameEndByLivingPlayers(out GameOverReason reason)
         {
             reason = GameOverReason.ImpostorByKill;
 
-            if (CustomRolesHelper.RoleExist(CustomRoles.Sunnyboy) && Main.AllAlivePlayerControls.Count() > 1) return false;
+            if (CustomRolesHelper.RoleExist(CustomRoles.Sunnyboy) && Main.AllAlivePlayerControls.Any()) return false;
```
Gurge44, line 494: This needs to be reverted cause count > 1 isn't the same as Any.
> 1 means 2 or more, while Any means 1 or more

dw I ran into the same problem 🤣

0xDrMoe: Could we not just do >= 1?

Gurge44: No cause it has to check if there's anyone else other than the sunnyboy alive, aka the player count is higher than 1 or not

Tommy-XL: Yes, I'll fix it, I didn't pay attention to it at the time

Tommy-XL: Fixed

Gurge44 (approved):

0xDrMoe:
ETA on this @Tommy-XL ?

Tommy-XL:
Well, this PR can be leaked right now, if necessary

Tommy-XL:
I won't have much time today or tomorrow to finish this, but I'll call it part 1. If you need to merge this as soon as possible then you can do it right now.  And when I have time, I'll just open a new PR and call it part 2 (if the current PR will be merged)

PR #106, opened 2023-10-21
Improve overkiller

NikoCat233:
Attempt to prevent target being spam killed.
Send rpc calls to alive players except target should be working
but this may cause huge rpc workloads for host
(IDK whether this will kick host, I cannot tested it in online games)

Note that originally client can only see bodies lying together at most conditions, so this pr does not change any thing else.

### DIFF
--- a/Modules/ExtendedPlayerControl.cs
+++ b/Modules/ExtendedPlayerControl.cs
@@ -266,16 +266,17 @@ static class ExtendedPlayerControl
         }
         player.ResetKillCooldown();
     }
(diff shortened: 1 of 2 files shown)


NikoCat233:
This works fine on local games

https://github.com/0xDrMoe/TownofHost-Enhanced/assets/139348239/7838b606-b3e6-44f3-ab23-d86fd8411e3f

NikoCat233:
Im adding something tomorrow

PR #107, opened 2023-10-21
Fix mini bug

LezaiYa:


Patches/PlayerControlPatch.cs
```diff
@@ -2770,7 +2770,7 @@ public static void Postfix(PlayerControl __instance)
                                         Logger.Info($"年龄增加1", "Child");
                                         if (Mini.UpDateAge.GetBool())
                                         {
-                                            player.Notify(GetString("MiniUp"));
+                                            if (player.Is(CustomRoles.NiceMini)) player.Notify(GetString("MiniUp"));
```
LezaiYa, line 2773: @Tommy-XL Why remove _if (player.Is(CustomRoles.Nice/Evil Mini))_    ?

Tommy-XL: Because this code is already in `case CustomRoles.NiceMini:` so I didn't see the point in additional checking

LezaiYa: Without the extra check, it will cause the player to not see the mini's age

LezaiYa: It's a bug that shouldn't be there ._.

0xDrMoe (approved):

Tommy-XL (approved):

PR #108, opened 2023-10-21
rework check game end patch

ryuk201198:
Tried to simplify a lot, needs A LOT of testing

All the contributors please look into the code and recommend changes.


0xDrMoe (approved):

Patches/CheckGameEndPatch.cs
```diff
@@ -492,61 +492,28 @@ public bool CheckGameEndByLivingPlayers(out GameOverReason reason)
...
+                if (pc == null) continue;
+
+                dual = pc.Is(CustomRoles.DualPersonality) ? 1 : 0;
+                var countType = Main.PlayerStates[pc.PlayerId].countTypes;
+                if (countType == CountTypes.OutOfGame || countType == CountTypes.None || countType == CountTypes.NWitch) continue;
+
+                if (roleCounts.ContainsKey(countType))
+                    roleCounts[countType]++;
+                else
+                    roleCounts[countType] = 1;
```
Gurge44, line 511: I'm tired so I may be wrong but.... if the dictionary doesn't contain the key (goes to else) and you still try to read that key, won't it just throw a KeyNotFoundException?
You'd need to add the key with a value of 1 to the dictionary using .Add or .TryAdd

Patches/CheckGameEndPatch.cs
```diff
@@ -557,231 +524,43 @@ public bool CheckGameEndByLivingPlayers(out GameOverReason reason)
...
                     {
+                        var winnerRole = winners[0].Key.GetNeutralCustomRoleFromCountType();
                         reason = GameOverReason.ImpostorByKill;
-                        winner = CustomWinner.Glitch;
-                        rl = CustomRoles.Glitch;
+                        CustomWinnerHolder.ResetAndSetWinner(CustomRolesHelper.GetNeutralCustomWinnerFromRole(winnerRole));
+                        CustomWinnerHolder.WinnerRoles.Add(winnerRole);
                     }
-                    else if (Jinx == totalNKAlive)
+                    else if (winners.Count == 0)
```
Gurge44, line 560: Use !winners.Any()

Patches/CheckGameEndPatch.cs
ryuk201198: Nope, you can add a key this way too

Gurge44 (approved):

Patches/CheckGameEndPatch.cs
```diff
@@ -557,231 +524,43 @@ public bool CheckGameEndByLivingPlayers(out GameOverReason reason)
 
             else if (totalNKAlive == 0) // total number of nks alive 0
             {
-                if (Crew == 0 && Imp == 0) // crew and Imp both 0 everyone dead
+                if (roleCounts[CountTypes.Crew] == 0 && roleCounts[CountTypes.Impostor] == 0) // Crew and Imp both 0, everyone is dead
```
Tommy-XL, line 527: Since `roleCounts[CountTypes.Impostor]` and `roleCounts[CountTypes.Crew]` are frequently used in the code, it would be helpful to create temporary variables for them, similar to `totalNKAlive`. These variables can be named `impCount` and `crewCount`, and they can be placed on line 523 - 524 of the code.

Tommy-XL: I mean remember this data and use it in code so as not to have to access the Dictionary very often

ryuk201198:
@Tommy-XL 
> I mean remember this data and use it in code so as not to have to access the Dictionary very often
What do you think of this?

@Gurge44 
> Use !winners.Any()
Since i was counting number of winners any way in the condition before, I used a variable to check the count so that i dont have to go through it again. What do you think?

Patches/CheckGameEndPatch.cs
```diff
@@ -557,231 +538,44 @@ public bool CheckGameEndByLivingPlayers(out GameOverReason reason)
...
-                    else if (Glitch == totalNKAlive)
+                    var winners = neutralRoleCounts.Where(kvp => kvp.Value == totalNKAlive).ToList();
+                    var winnnerLength = winners.Count;
+                    if (winnnerLength == 1)
                     {
+                        var winnerRole = winners[0].Key.GetNeutralCustomRoleFromCountType();
                         reason = GameOverReason.ImpostorByKill;
-                        winner = CustomWinner.Glitch;
-                        rl = CustomRoles.Glitch;
+                        CustomWinnerHolder.ResetAndSetWinner(CustomRolesHelper.GetNeutralCustomWinnerFromRole(winnerRole));
```
Tommy-XL, line 572: Use `CustomWinnerHolder.ResetAndSetWinner(winnerRole.GetNeutralCustomWinnerFromRole())` because the function has `this` method

Tommy-XL: Also add `using TOHE.Modules;`

Tommy-XL (approved):

PR #117, opened 2023-10-24
fixed recruiting roles showing up twice in end screen

ryuk201198:
fix buggy end screen with reworked game end and merged with main


Tommy-XL:
Hmm, I think this code will cause a bug for Vanilla players when the end screen is just blank, but I'm not sure about that

ryuk201198:
> Hmm, I think this code will cause a bug for Vanilla players when the end screen is just blank, but I'm not sure about that

let me test it real quick.

ryuk201198:
@Tommy-XL just tested it with cultist. It worked fine

Tommy-XL (approved):

PR #119, opened 2023-10-24
Fix Reset Settings?

Tommy-XL:


Tommy-XL:
As it turned out, this is a completely different bug and the settings are not reset due to `BoolOptionNames.GhostsDoTasks`, so I will cancel these changes later

PR #140, opened 2023-10-27
Fix mini bug

LezaiYa:
1.Fix mini being able to assign add-ons that shouldn't be assigned
2.Fix Version 10.24 causes a bug that prevents mini from growing up


NikoCat233:
. in game its evil mini and nice mini. U can't check addon conflict with mini
Also why delete syncminiage?
And "Fix Version 10.24 causes a bug that prevents mini from growing up" What is this
And u need to merge into updatenewversionAU instead of main

PR #144, opened 2023-10-28
Try to fix SabotageMaster and Alchemist

Gurge44:
If it works, just add the same for Repairman and it should work.


Tommy-XL:
Yes, it works for modded but not vanilla

PR #145, opened 2023-10-28
Add disable comms on fungle

Gurge44:


Tommy-XL:
I think this setting should be enabled by default, but I'm not sure everyone will like it

Gurge44:
I thought about it, but then I made it off by default since every other one was off by default

Tommy-XL (approved):

PR #160, opened 2023-10-29
Dev Manager role should show in every TOHE lobby :trollface: 

FuroYT:
(This is a small change but cool i think)

### DIFF
--- a/Modules/Utils.cs
+++ b/Modules/Utils.cs
@@ -1859,7 +1859,7 @@ public static class Utils
                 };
             }
             
-            if (!name.Contains('\r') && player.FriendCode.GetDevUser().HasTag())
+            if (!name.Contains('\r') && player.FriendCode.GetDevUser().HasTag() && !Main.AutoMuteUs.Value)
             {
                 name = player.FriendCode.GetDevUser().GetTag() + "<size=1.5>" + modtag + "</size>" + name;
             }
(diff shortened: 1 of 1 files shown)


NikoCat233:
This may cause some user's name get blocked or covered

FuroYT:
I tried on an TOHE v1.1.1 build and it works fine didn’t notice any problem

NikoCat233:
Its bad to find suffix got blocked by dev tags . this should be working

ryuk201198:
Long tags can cause mass disconnect issues. If dev tag is added, suffix was blocked to prevent from dc issues from happening.

FuroYT:
> Long tags can cause mass disconnect issues. If dev tag is added, suffix was blocked to prevent dc issues from happening.

I havent experienced any disconnect issue with this and I was playing since 2 weeks ago not any disconnect issue

ryuk201198:
length of the tag was the reason gradients were removed and/or made a setting

FuroYT:
> length of the tag was the reason gradients were removed and/or made a setting

The tag itself doesn’t change but it’s just the way that the dev manager tag is shown that changed here so the length disconnect issue isn’t due to this but the servers or game using too much colors itself

Modules/Utils.cs
```diff
@@ -1862,6 +1858,11 @@ public static void ApplySuffix(PlayerControl player)
                     _ => name
                 };
             }
+            
+            if (!name.Contains('\r') && player.FriendCode.GetDevUser().HasTag())
```
Tommy-XL, line 1862: Why `&& !Main.AutoMuteUs.Value`  was removed?

FuroYT: It’s because I changed the code before it was added

Tommy-XL (approved):

PR #180, opened 2023-10-30
Fix admirer, jackal, succubus, mario. Update string. Improve /changerole

NikoCat233:
1.Recode admirer for better performance
2.Fix admirer knowrole don't work properly
3.Fix admired mario win condition check

4.Fix jackal killing admirer, sheriff (supposed to recruit) when use only recruit
(To prevent jackal from recruiting neutrals, I use !isNeutral(). Why? No more Toonie admirer! )

5.Temporary prevent succubus from charming neutrals. (God damn who allowed this? I dont want to fix neutrals win condition check like Toonie admirer)

6.Improve command /changerole 
    a.(free in local game and limited to debug user on online game)
    b.Now it do not use contain to assign mutiple roles (Like "魅魔" in chinese , it assignes both "Charmed" and "succubus" together previously)

7.Finally update the string
8.Now If host is versioncheating, then mod clients wont automatically quit


NikoCat233:
Tested in online games, everyone works fine

Tommy-XL (approved):

PR #182, opened 2023-10-30
Fix taskbar always invisible

GayKevin:
Fix: Remove code that was hidding the taskbar instead of using the setting in game of the TaskBarMode


Tommy-XL (approved):
Loonie added this because taskbar does not correctly display the progress of tasks, although I have never noticed this

PR #198, opened 2023-11-02
auto copy build

PHGaming121:
auto copies the build into game directory
it got deleted, but im placing it again


PHGaming121:
also, something i forgot when translating cmds

PR #204, opened 2023-11-03
New role Taskinator and made mario incompatible with unlucky

ryuk201198:
Review please


Roles/Neutral/Taskinator.cs
```diff
@@ -0,0 +1,139 @@
...
+        taskIndex = new();
+        TaskMarkPerRound = new();
+        IsEnable = false;
+        maxTasksMarkedPerRound = TaskMarkPerRoundOpt.GetInt();
+    }
+    public static void Add(byte playerId)
+    {
+        playerIdList.Add(playerId);
+        TaskMarkPerRound[playerId] = 0;
+        IsEnable = true;
```
Tommy-XL, line 40: Seems `playerIdList` it's not used anywhere, so it can be removed

main.cs
```diff
@@ -1008,6 +1010,7 @@ public enum AdditionalWinners
     Phantom = CustomRoles.Phantom,
     Maverick = CustomRoles.Maverick,
     Shaman = CustomRoles.Shaman,
+    Taskinator = CustomRoles.Taskinator,
```
Tommy-XL, line 1013: You forgot add `AdditionalWinners.Taskinator` in `en_US.json`

Roles/Neutral/Taskinator.cs
ryuk201198: I plan on using it in the future, for a different role

Tommy-XL: Hmm, ok

main.cs
ryuk201198: done

Tommy-XL (approved):

PR #219, opened 2023-11-05
New role Pixie, taskinator RPC fix 

ryuk201198:
I've also changed the Id comment in OptionHolder.cs

Also please Review


Limeau:
Pixie role? :O what does it do?

PR #224, opened 2023-11-05
Option to hide Shapeshift Animation 

lars-wu:
Option for Dazzler, Devourer, Deathpact and Pitfall to hide the Shapeshift Animation for their Abilities.

I implemented this after reading @Gurge44's comment.
https://discord.com/channels/1094344790910455908/1159602185353580644/1169640524186136707


Tommy-XL:
I was going to do this too, but as a general setting for all roles with the ability Shapeshift

Gurge44:
Yea and you can also do it for a lot more roles than just these 4. But note that by doing this, you cancel the shapeshift completely, not just the animation. The player will not transform at all. You can see how I've done it in TOHE+.

Dawson2763:
What about Escapist?

Gurge44:
And Miner, Disperser, Evil Tracker, Bomber, Nuker, Quick Shooter, and a lot more roles

lars-wu:
This does not work for vanilla clients unfortunately. After using the shapeshift ability they get stuck and are unable do anything. 
I also tested it in TOHE+ with the same result. Maybe I'm missing something, but seems like it just does not work.

Tommy-XL:
Bad

Gurge44:
I fixed it in my latest commit

Ultradragon005:
I call cap
### **_**A**_**

0xDrMoe:
whats happening here?

lars-wu:
forgot to close it.

PR #254, opened 2023-11-15
Update EAC

NikoCat233:
Add sabotage check for EAC
1. player with crew basis can not sabotage
2. player who send rpc to directly sabotage (amount 128 , 16, electric > 50) will be checked if it has pre send systemtypes.sabotage and agreed by host

Tommy can you pls help me with stuffs in sabotage patch?
I return false in shipstatuspatch but sabotage is still triggered by malummenu
probably bcz it does not send systemtypes sabotage for host to check but amount 128 to directly sabotage

### DIFF
--- a/Modules/ChatManager.cs
+++ b/Modules/ChatManager.cs
@@ -191,7 +191,6 @@ namespace TOHE.Modules.ChatManager
                 {
                     senderPlayer.Revive();
                 }
-
                 DestroyableSingleton<HudManager>.Instance.Chat.AddChat(senderPlayer, senderMessage);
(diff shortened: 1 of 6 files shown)


NikoCat233:
will antiblackout for Reset cam player trigger eac?
pls help me check it

NikoCat233:
Tested in my online public lobby for 4 hours and everything seems fine.
Suggest a dev build to further test this

Tommy-XL:
I think it's better to merge this after the 1.2.1 release

NikoCat233:
Ok

PR #258, opened 2023-11-18
Make convert more superb

NikoCat233:
1.Make jackal and succubus recruit players like admirer and gangster
2.Put all these checks into one manager

3.Add setting : superb convert
if a recruiter get recruited into another team, then all the player recruited by the recruiter will also join that new team

4.Improve opportunitist immue to kill && reset tasks for those who turns to opportunitist in game
5.Sometimes a avanger avanges another avanger and causes a never ending loop. Try if set isdead works for this

### DIFF
--- a/Modules/ConvertManager.cs
+++ b/Modules/ConvertManager.cs
@@ -1,4 +1,5 @@
 ﻿using Hazel;
+using LibCpp2IL.Elf;
 using System.Collections.Generic;
 using System.Linq;
 using TOHE.Roles.AddOns.Crewmate;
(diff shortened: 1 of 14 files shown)


NikoCat233:
Oops, i find knworoletarget extremely hard to do if we allow role like jackal to recruit player into other team

NikoCat233:
Again tested in my online public lobbies for 1 hour with superb convert enabled
Should be working fine

Currently only support gangster, admirer and jackal
Know role target does not allow 1th recruiter to see 3th recruited

NikoCat233:
Leaving this for next dev version

PR #265, opened 2023-11-19
Added /rand

MargaretTheFool:
/rand gives out a random number between two numbers inputted by the user (including both numbers)

suggested by TronAndRey


ryuk201198:
@MargaretTheFool Why in the world is there a new ChatCommandPatch.cs file??

MargaretTheFool:
i uploaded it in the wrong place

MargaretTheFool:
which i thought I already deleted, but I guess not

Tommy-XL (approved):

PR #275, opened 2023-11-24
added new crew role Mole

ryuk201198:


ryuk201198:
Merge this after the full release

ryuk201198:
Please review

Tommy-XL (approved):

Roles/Crewmate/Mole.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Linq;
+using UnityEngine;
+using static TOHE.Options;
+
+namespace TOHE.Roles.Crewmate
+{
+    public static class Mole
+    {
+        private static readonly int Id = 26000;
+        private static List<byte> playerIdList = new();
```
Tommy-XL, line 11: I think that `playerIdList` can be removed, since it is not used in the code

ryuk201198: > I think that `playerIdList` can be removed, since it is not used in the code

done

PR #278, opened 2023-11-25
Fix OIIAI neutral assign

NikoCat233:
Bug: 
Neutrals but not NK will be erased to Crewmate

Solution:
Add function HasImpKillButton to decide whether the player has kill button
if they do, use NK assign
If they dont, assign opportunitist

Also reset opportunitist tasks upon assigning


NikoCat233:
HasImpKillButton is also used to decide whether to reset the kcd of players

Tommy-XL (approved):

PR #283, opened 2023-11-28
Ported FFA from TOHE+ (Thanks gurge)

ryuk201198:


GameModes/FFAManager.cs
```diff
@@ -0,0 +1,444 @@
...
+        originalSpeed = new();
+        KBScore = new();
+        RoundTime = FFA_GameTime.GetInt() + 8;
+
+        foreach (PlayerControl pc in Main.AllAlivePlayerControls)
+        {
+            KBScore.TryAdd(pc.PlayerId, 0);
+        }
+
+        _ = new LateTask(Utils.SetChatVisible, 10f, "Set Chat Visible for Everyone");
```
Gurge44, line 101: fyi I just tested it and it also works with 4f delay which doesn't interrupt as much, so you can change it to that if you want to

Resources/Lang/en_US.json
```diff
@@ -200,6 +200,7 @@
   "Investigator": "Investigator",
   "Guardian": "Guardian",
   "Addict": "Addict",
+  "Mole": "Mole",
```
Tommy-XL, line 203: It looks like there is a Mole code here that has already been done in #275

ryuk201198: Yeah, the base of the branch was the mole branch.
I'll fix any conflicts that may arrise after the mole PR has been merged

0xDrMoe:
Can we get this done n' ready?

ryuk201198:
Yeah, i'll be working on it.. there are some issues that needs to be taken care of.

ryuk201198:
Please Review

Tommy-XL (approved):

Tommy-XL:
@Gurge44 Will be better if you check the code too, I just don't quite understand this mode, so I could have missed something

Gurge44:
> @Gurge44 Will be better if you check the code too, I just don't quite understand this mode, so I could have missed something

It's simple: Everyone is a basic neutral killer with no special ability, everyone can kill everyone, the last player alive wins. To spice this up, after each kill, the killer gets a random effect, like a lower kill cooldown or a speed boost. Depending on the settings, everyone may get swapped with someone from time to time. The last 2 players alive will get an arrow pointing towards each other.

Gurge44 (approved):
Apart from the fact that the kcd timers could've been solved with less than 3 dictionaries, it should be good by first look.

Tommy-XL:
Well, then it will be merged

PR #301, opened 2023-12-02
Delay send message  for invalid color player

NikoCat233:
Seems useless


Tommy-XL:
I don't quite understand why this is needed

NikoCat233:
I found that when a player join with invalid color, after send message, the host freaking disconnect cause "Reliable packet not receive"

NikoCat233:
Im not sure of the revelance

PR #308, opened 2023-12-03
fix bug and some change

ryuk201198:
1. added extra delay to chat visible code to solve chat not showing because of high ping issue
2. Only allow people with debugger perms to turn on console (people will still see the option but if they turn on, nothing will happen)


ryuk201198:
I've done this because normal players should not see the console. It will also help in stopping players from console sniping

PR #311, opened 2023-12-04
Update EAC

NikoCat233:
1.Sabotage check "Directly patches rpc"
2.Rework report dead body check
3.Other obvious and certain cheats check
4.Add tempban option to eac handle cheat

Hosted in online public games without protocol for 1 hour and seems fine
Not sure whether using version protocol, clients will no longer directly send murder player rpc, I did not test it with protocol


NikoCat233:
Hard to believe that I forget to reopen the pr

PR #317, opened 2023-12-06
made few roles unable to spawn in dlekS

ryuk201198:
made arsonist, spellcaster and hexmaster unable to spawn if switch mode is set to vent

### DIFF
--- a/Modules/CustomRoleSelector.cs
+++ b/Modules/CustomRoleSelector.cs
@@ -95,7 +95,10 @@ internal class CustomRoleSelector
                     or CustomRoles.Lighter
                     or CustomRoles.Chameleon
                     or CustomRoles.Mario // Vector
-                    or CustomRoles.Wraith) continue;
+                    or CustomRoles.Wraith
+                    or CustomRoles.Arsonist) continue;
+                if (role == CustomRoles.Witch && (Witch.SwitchTrigger)Witch.ModeSwitchAction.GetValue() == Witch.SwitchTrigger.Vent || // Spellcaster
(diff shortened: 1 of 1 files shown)


Modules/CustomRoleSelector.cs
```diff
@@ -95,7 +95,10 @@ or CustomRoles.TimeMaster
                     or CustomRoles.Lighter
                     or CustomRoles.Chameleon
                     or CustomRoles.Mario // Vector
-                    or CustomRoles.Wraith) continue;
+                    or CustomRoles.Wraith
+                    or CustomRoles.Arsonist) continue;
```
Tommy-XL, line 99: Arsonist was been fixed on this map so he can appear on this map

ryuk201198: wdym? arsonist needs to vent after dousing everyone, right?

Tommy-XL: Yes, but only once, when all players are doused

ryuk201198: so how is arsonist going to vent?
Also what about the setting where it can vent to kill set amount of players?

Tommy-XL: Vents is only partially broken, but not completely

Tommy-XL: > so how is arsonist going to vent? Also what about the setting where it can vent to kill set amount of players?

Damn, I forgot he can do that :/

Tommy-XL (approved):

Modules/CustomRoleSelector.cs
Tommy-XL: Fix it later

PR #328, opened 2023-12-11
Revert changes in HGK

NikoCat233:
1. Sorry i misunderstood shieldpersondeadfirst
2. Comment out EAC check name bcz its broken
3. Change result flags in rpc guard kill bcz FailedProtected is what vanilla does and its much safer


NikoCat233:
Note that playerid and client id may change when a player rejoin the room
So use hash puid

Tommy-XL (approved):

PR #329, opened 2023-12-12
Improve meeting hud check

NikoCat233:
Patch CastVote and do rpcclearvote there

1. Patch dictator vote solsticer
2. Patch player from voting a dead player (guessed), and should return the vote if target is dead or null
3. Improve the logic to skip vote finished checks


NikoCat233:
Forget to return true in cast vote for non host
help me add it pls

Tommy-XL (approved):

PR #342, opened 2023-12-15
Addon Silent

NikoCat233:
EZ code
changed repairman namespace to addons.common

PR #344, opened 2023-12-16
New addon Mundane

ryuk201198:


NikoCat233:
Ryuk are u crazy...

ryuk201198:
> Ryuk are u crazy... Imagine 5 roles in a week

lmao have to open up the role ideas to everyone, It has been too long since it has been restricted for few people

Roles/AddOns/Common/Mundane.cs
```diff
@@ -0,0 +1,25 @@
+﻿using static TOHE.Options;
+
+namespace TOHE.Roles.AddOns.Common
+{
+    public static class Mundane
+    {
+        private static readonly int Id = 26600;
```
Tommy-XL, line 7: Need to change to 26700 since 26600 is already used in #342

PR #346, opened 2023-12-16
new role guess master and added doomsayer and solsticer to divinator check list

ryuk201198:
Please Review

merge after #348


Roles/Crewmate/GuessMaster.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Collections.Generic;
+using static TOHE.Translator;
+
+namespace TOHE.Roles.Crewmate
+{
+    public static class GuessMaster
+    {
+        private static readonly int Id = 26700;
```
Tommy-XL, line 8: Well, same need to change to 26800 since 26700 is already will be used in #344

PR #376, opened 2023-12-24
Add Kamikaze

Ultradragon005:
brb have to also add the other files


Ultradragon005:
Please review

LezaiYa:
Please wait for the next development before adding new roles.

NikoCat233:
Also pls add your role in divatator results

Roles/Impostor/Kamikaze.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        
+        var KamiId = reader.ReadByte();
+        var targetId = reader.ReadByte();
+        byte KillerIdo = reader.ReadByte();
+        int Limit = reader.ReadInt32();
+        
+        KamikazedList[targetId] = KamiId;
+        MarkedLim.Add(KillerIdo, Limit);
+        MarkedLim[KillerIdo] = Limit;
```
Gurge44, line 76: You can remove the .Add statement as the statement after it does the same but without exceptions

Roles/Impostor/Kamikaze.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            KamikazedList.Remove(kamikameha.PlayerId);
+        }
+        else if(CheckKamiDeath)
+        {
+            foreach (var alivePlayer in Main.AllAlivePlayerControls)
+            {
+                Main.PlayerStates[kamikameha.PlayerId].deathReason = PlayerState.DeathReason.Targeted;
+                alivePlayer.RpcMurderPlayerV3(kamikameha);
+            }
```
Gurge44, line 93: Why do all alive players murder this one single player?!

Patches/PlayerControlPatch.cs
```diff
@@ -2504,6 +2507,8 @@ public static void Postfix(PlayerControl __instance)
             LocateArrow.OnFixedUpdate(player);
         }
 
+        if (Kamikaze.IsEnable)
+            Kamikaze.MurderKamikazedPlayers(player);
```
Gurge44, line 2511: As this isn't a priority, put this into`if (!lowLoad)` and this should only run on the host's side, so move it into `if (AmongUsClient.Instance.AmHost)`

Roles/Impostor/Kamikaze.cs
```diff
@@ -0,0 +1,130 @@
...
+using UnityEngine;
+using static TOHE.Options;
+using static TOHE.Translator;
+
+namespace TOHE.Roles.Impostor;
+
+public static class Kamikaze
+{
+    private static readonly int Id = 26900;
+    public static List<byte> MaxMarked = new();
```
Gurge44, line 15: I couldn't find any use for this list, what is the purpose of it exactly?

Roles/Impostor/Kamikaze.cs
Ultradragon005: OH RIGHT, I forgot to check if aliveplayer corresponds with kamikameha player. Fixed now.

0xDrMoe:
Is this ready?

Ultradragon005:
> Is this ready?

yup!

PR #377, opened 2023-12-25
Auto Rehost for debugger Usage

NikoCat233:
This feature may be back to InnerSloth's Policy
So only limit to debug usage

Only functions as host disconnect and not ExitGame
Use keybind LeftShift + C to cancel a rehost countdown
Also U can cancel host side auto play again with LeftShift + C

### DIFF
--- a/Patches/ClientOptionsPatch.cs
+++ b/Patches/ClientOptionsPatch.cs
@@ -23,8 +23,9 @@ public static class OptionsMenuBehaviourStartPatch
     private static ClientOptionItem SwitchVanilla;
 #if DEBUG
     private static ClientOptionItem VersionCheat;
+    private static ClientOptionItem GodMode;
+    private static ClientOptionItem AutoRehost;
 #endif
-    //private static ClientOptionItem GodMode;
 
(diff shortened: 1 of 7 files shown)


Ultradragon005:
mega OP OP

Tommy-XL:
I think it's better to leave this for the next dev version

Tommy-XL (approved):

0xDrMoe:
Is this ready?

Patches/ClientOptionsPatch.cs
```diff
@@ -23,8 +23,9 @@ public static class OptionsMenuBehaviourStartPatch
     private static ClientOptionItem SwitchVanilla;
 #if DEBUG
     private static ClientOptionItem VersionCheat;
+    private static ClientOptionItem GodMode;
+    private static ClientOptionItem AutoRehost;
 #endif
```
Tommy-XL, line 28: If dev versions are built as debug, then these settings will appear for users of the dev version

NikoCat233: I dont think we are building debug...

PR #378, opened 2023-12-27
Change flags in rpcguardandkill

NikoCat233:
To declare to client that this call is sent by host


Tommy-XL (approved):

NikoCat233:
WTF you are too fast

PR #383, opened 2023-12-30
Encode role log and bug fix

hs233666:
1. encode role log with DES
2. /dump now will be logged by host with rpc calls

3. Fixed /v sometimes not working


Tommy-XL:
I think merge this after the release version

0xDrMoe:
I disagree @Tommy-XL I think it should be in this version

Tommy-XL (approved):

PR #399, opened 2024-01-03
Support Vanilla Hide&Seek & Fix Vents On Dleks Map

Tommy-XL:
1. Support Vanilla Hide&Seek
2. Fixed Vents On Dleks Map (For Modded)


Modules/TemplateManager.cs
```diff
@@ -119,7 +134,11 @@ public static void SendTemplate(string str = "", byte playerId = 0xff, bool noEr
             playerName = () => Main.AllPlayerNames[playerId];   
         }
 
-        _replaceDictionary["PlayerName"] = playerName;
+        //if (GameStates.IsNormalGame)
+            _replaceDictionaryNormalOptions["PlayerName"] = playerName;
+        //else if (GameStates.IsHideNSeek)
+            _replaceDictionaryHideNSeekOptions["PlayerName"] = playerName;
+
```
Tommy-XL, line 141: For some reason, when creating a lobby, `"PlayerName"` is null
Need to find a solution to the problem

Tommy-XL:
399 PR, Pog

Modules/TemplateManager.cs
Tommy-XL: I found out that this only happens when creating a local lobby, I don’t know why, lol

Tommy-XL:
I can say that at this stage H&S is at the testing stage

Tommy-XL:
> What roles do we support in hide n seek?

I don't plan to support any roles in H&S at all, Just some settings that work well in H&S

Tommy-XL:
like Decon time, Random maps, Disable ZipLine etc.

Modules/TemplateManager.cs
Tommy-XL: Fixed

NikoCat233:
> > like Decon time, Random maps, Disable ZipLine etc.
> 
> can you straight up disable decon, like make both doors permanently open

This isnt possible.
IS dont provide a flag to open both door

0xDrMoe:
ETA on this?

Tommy-XL:
I think everything is ready

Tommy-XL:
> I think everything is ready

Nope, lol

Tommy-XL:
Now done
(Also in the future need to fix random spawn as the positions of the cards have been changed)

PR #411, opened 2024-01-06
Fuck up Version Check

NikoCat233:
Change version check from using playerId to using client Id
Bcz client id is unique

Fix On Host Disconnect action not working
How superising! This bug exists since tohe 2.3.6


0xDrMoe:
Ready to pull?

NikoCat233:
Ready

PR #416, opened 2024-01-07
Add option to hide host text + Translate API tags

Pietrodjaowjao:


Pietrodjaowjao:
Please review, you can remove the pt-br translation so it wont bug crowdin.

Pietrodjaowjao:
ig it can be merged now

Pietrodjaowjao:
> ig it can be merged now

@Tommy-XL

0xDrMoe:
Is this automatic translation? I don't think it should be automatic because some translators don't want it to be in their language. They want just English.

Pietrodjaowjao:
> Is this automatic translation? I don't think it should be automatic because some translators don't want it to be in their language. They want just English.

Nope, i'm brazilian so i know exactly what to put, there's also a error in the original pt-br translation.

Sponsor for some reason is 'Tradutor', it should be 'Patrocinador'

Pietrodjaowjao:
If it doens't bug crowdin, i can put the translations again if you want

Pietrodjaowjao:
If the translator don't want it to be their languague, they can simply leave the original text in translations file.

PR #422, opened 2024-01-07
Fix Check Game End & Some Fix

Tommy-XL:


Patches/CheckGameEndPatch.cs
```diff
@@ -612,11 +610,14 @@ public bool CheckGameEndByLivingPlayers(out GameOverReason reason)
...
+
                 else if (impCount == 0) // Remaining Imps are 0, Crew wins (neutral is already dead)
                 {
                     reason = GameOverReason.HumansByVote;
                     CustomWinnerHolder.ResetAndSetWinner(CustomWinner.Crewmate);
                 }
+
+                else if (crewCount > impCount) return false;
                 return true;
             }
```
Tommy-XL, line 622: When Crewmate turned out to be larger than Impostors and there were no specified reasons for end game, it always returned true, which could possibly cause many bugs with the end of the game

PR #425, opened 2024-01-08
Optional bool to use vanilla strings.

Pietrodjaowjao:
always use vanilla:yourBool to make sure it wont make other thing true


Pietrodjaowjao:
This repo is completely broken. will be creating another one with a better way to do that

PR #426, opened 2024-01-08
Optional bool to use vanilla strings. (fixed)

Pietrodjaowjao:
This lets you use vanilla strings on translator and on GameOptions.


Pietrodjaowjao:
hmm

Tommy-XL:
I don't know where this can be used other than Num Impostors

Tommy-XL (approved):

PR #430, opened 2024-01-09
Add Soloist

Limeau:
**Soloist** - a neutral role where you must kill everyone to win


Limeau:
Okay wtf I tried to put this on the TOB fork

PR #463, opened 2024-01-14
make /rand inclusive + add /rn for vanilla clients

MargaretTheFool:
i lied and said it was inclusive in the string for it but it wasnt


MargaretTheFool:
ok i DONT know why my commits for something else went into this too

0xDrMoe:
is this ready if /rn is here?

Patches/ChatCommandPatch.cs
```diff
@@ -1482,6 +1482,16 @@ public static void OnReceiveChat(PlayerControl player, string text, out bool can
...
+            case "/rn":
+            case "/rename":
+            case "/renomear":
+                if (args.Length < 1) break;
+                if (args.Skip(1).Join(delimiter: " ").Length is > 10 or < 1) {
+                    Utils.SendMessage(GetString("Message.AllowNameLength"), player.PlayerId);
+                    break;
+                }
+                Main.AllPlayerNames[player.PlayerId] = args.Skip(1).Join(delimiter: " ");
+                break;
```
Tommy-XL, line 1494: Wouldn't it be better to do this as a setting?

Tommy-XL: For example "Vanilla players can use command /rename"

MargaretTheFool:
ok NOW everything should be ready

PR #465, opened 2024-01-15
fix /rand and /gno giving same number

ryuk201198:


MargaretTheFool:
#463 i just fixed it :hehe:

PR #494, opened 2024-01-20
Rebase Swapper

NikoCat233:
1. Fix swapper votes sometimes arent swapped
2. Fix swapper skill limit arent decreased after meeting and swapper can swap freely
3. Fix swapper skill count arent synced to clients
4. Fix mutiple swapper in a round will bug the votes
5. Improve swapper notify
6. Now you can use /sw 253 to clear your swap selections


NikoCat233:
It would be funny if we have mutiple swapper and they swap a same target
At least the code wont bug

Tommy-XL (approved):

PR #512, opened 2024-01-21
Port some TOH (taken from TOHE+) roles.

Ultradragon005:
- [ ] Chek if stealth works fine.
- [ ] Check if penguin works fine.
- [ ] Check If PlagueDoctor (Plague Scientist) works fine.


Ultradragon005:
Should penguin be in experimental because of rpcteleport spam?

PR #517, opened 2024-01-22
add new role schrodingers cat

ryuk201198:


NikoCat233:
how many cats do we have now in tohe lol

Ultradragon005:
> how many cats do we have now in tohe lol

WAYYYY too little, and so, I shall rename https://discord.com/channels/1094344790910455908/1180573554476060832 to "Scaredycat"

Resources/roleColor.json
Ultradragon005, line 1: at this point why not simply leave it at such spacing, VS keeps auto-formatting it to this anyways..

PR #531, opened 2024-01-25
Patch check shapeshit

NikoCat233:
Patch checkshapeshift
so some roles with 1 ss duration wont have to suffer the pain of useless shape shift

### DIFF
--- a/Roles/Impostor/RiftMaker.cs
+++ b/Roles/Impostor/RiftMaker.cs
@@ -113,7 +113,8 @@ public static class RiftMaker
     public static void ApplyGameOptions()
     {
         AURoleOptions.ShapeshifterCooldown = SSCooldown.GetFloat();
-        AURoleOptions.ShapeshifterLeaveSkin = true;
+        //AURoleOptions.ShapeshifterLeaveSkin = true;
+        //You set leave skin here only influences the rifter it own. Its not global
         AURoleOptions.ShapeshifterDuration = 1f;
     }
     public static void SetKillCooldown(byte id) => Main.AllPlayerKillCooldown[id] = KillCooldown.GetFloat();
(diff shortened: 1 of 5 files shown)


NikoCat233:
Gonna support the roles i changed in part 1 first.
Some roles have their own ss duration so I will leave it later.

Roles/Impostor/RiftMaker.cs
```diff
@@ -113,7 +113,8 @@ public static void ReceiveRPC(MessageReader reader)
     public static void ApplyGameOptions()
     {
         AURoleOptions.ShapeshifterCooldown = SSCooldown.GetFloat();
-        AURoleOptions.ShapeshifterLeaveSkin = true;
+        //AURoleOptions.ShapeshifterLeaveSkin = true;
+        //You set leave skin here only influences the rifter it own. Its not global
         AURoleOptions.ShapeshifterDuration = 1f;
```
Tommy-XL, line 118: What? It should be because for Rift Maker its helps a lot to determine exactly where he placed the marker.

NikoCat233: huh, but if we reject the shapeshift animation, there wont be skin left

Tommy-XL: Bad

NikoCat233:
Currently these roles' shapeshift animation is hidden:
Undertaker, EvilTracker, Hacker, Bomber, Nuker, QuickShooter, Dazzler, Devourer, Blackmailer, RiftMaker, Pitfall

these roles' shapeshift calls is rejected bcz their ss button is used to display skill timer:
Penguin, BountyHunter, SerialKiller

Roles that have their own shapeshift duration settings will not be reworked in this pr.

We need to make sure the roles mentioned above works fine with the reject shapeshift technique first, then we can rework more roles. Pls roll out a build to test the roles above.

Tommy-XL:
I think it's better to leave this for 1.6.0 (after release)

NikoCat233:
OK

Ultradragon005:
This works for vanilla? I remember something else #224 like this being made and #224 didn't work for vanilla

Tommy-XL:
Should be working
image

Tommy-XL:
`player.RpcRejectShapeshift()` As far as I understand, it completely cancels the effect of shapeshift

Ultradragon005:
yeah I just noticed that previous pr I mentione didn't use that.. so great then

Tommy-XL:
Yeah, vanilla players froze in place because `player.waitingForShapeshiftResponse` return `true`
image

Gurge44:
I've done this months ago in TOHE+ and it works perfectly.
RejectShapeShift just sets waitingForShapeshiftResponse to false

PR #541, opened 2024-01-26
Move versioncheck to pc.onenable

NikoCat233:
Also added a shouldLog bool in latetask


NikoCat233:
Ready to merge i think

PR #558, opened 2024-01-28
Make FixedUpdatePatch Async

Tommy-XL:


Tommy-XL:
FixedUpdatePatch now will be processed async (may improve performance on slower devices)

Tommy-XL:
Btw, the idea was taken from [TOHE+](https://github.com/Gurge44/TOHE_PLUS)

PR #606, opened 2024-02-04
Add a tag for the translation of the website

LezaiYa1:
Wait for moe to agree, if not agree will close this PR


0xDrMoe:
I'd rather have this mitigated completely from the code

Ultradragon005:
since when do chinese people get special tags there's an entire section for specific chinese people

LezaiYa1:
> since when do chinese people get special tags there's an entire section for specific chinese people

TOHE's mod was first made by KARPED1EM, and then it was passed down, and KARPED1EM is Chinese, and the tags for China were added by KARPED1EM, not by us.

LezaiYa1:
> I'd rather have this mitigated completely from the code

What do you mean?

NikoCat233:
意思是他会通过api加，你在discord上面发给他

LezaiYa1:
> 意思是他会通过api加，你在discord上面发给他

我已经开票了啊（

PR #609, opened 2024-02-04
Option to disable votekick system.

Pietrodjaowjao:


Pietrodjaowjao:
This will prevent hackers like Silatiscm to kick ppl

PR #621, opened 2024-02-07
Recovering BanWords kicks out

LezaiYa1:


NikoCat233:
Need to exclude the condition where host uses other player's pc to send chats when host is dead

PR #628, opened 2024-02-10
Fix mod updater for current version number

NikoCat233:
Make ModUpdater directly download dll instead of zip
Force updater and dbconnect to try to use system proxy (if exists)
Add detailed log for dbconnect failed response

Make ModUpdater nolonger use async to avoid crashes (Thanks to Pietro)

*** For further Release ***
The release tag must be v{version_number}
eg v2024.xxxx.xxxx.xxxx
Tags like v1.5.1 will result in wrong result
And it must contains a TOHE.dll for mod updater to download


NikoCat233:
Gonna to discuss details with our devs

Tommy-XL:
It's done?

NikoCat233:
Its done, but possibly i need to talk with moe on version number thing and add some additional code

Tommy-XL:
Now it's done?

NikoCat233:
Yep

PR #648, opened 2024-02-15
Change delay of check murder && Add LateKill for player on ladder

NikoCat233:


Patches/PlayerControlPatch.cs
```diff
@@ -128,6 +128,7 @@ public static bool Prefix(PlayerControl __instance, [HarmonyArgument(0)] PlayerC
 
         // Is the target in a killable state?
         if (target.Data == null // Check if PlayerData is not null
+            || target.PlayerId == killer.PlayerId // Impossible interaction
             // Check target status
```
Tommy-XL, line 132: I think Sniper code says otherwise

image

NikoCat233: Then it should not use check murder

NikoCat233:
GG this pr

PR #651, opened 2024-02-15
Remove DecisionByHost

NikoCat233:
DecisionByHost is 100% useless and will cause trouble bcz it let client locally check protection
Also patch cmdCheckMurder, it exactly works as the code in the patch


NikoCat233:
Plan to add a way to detect server side version protocol

NikoCat233:
Welp as long as we are getting rid of DecisionByHost, its ok

PR #662, opened 2024-02-17
Fix Time Master teleport & Random spawn

Tommy-XL:
~~Seems `lastSequenceId` is sync with a large delay between the host and vanilla players, because of this the Host stands still for several seconds after teleport (for vanilla) while the host sees everything normally~~


Tommy-XL:
I want to test PR tomorrow

PR #671, opened 2024-02-18
Fix issues in ghost assign

Ultradragon005:


Ultradragon005:
Fixed, OptionItem was constantly refreshing so I made an Init task.
Also moved it to a seperate file bcuz it was getting humongous.

Tommy-XL (approved):

PR #677, opened 2024-02-19
Hotfix

Ultradragon005:
always assign with amhost, ignore players who don't exist. Debug & /Up unable to assign ghost-roles bcuz it breaks the game.


Ultradragon005:
okay so I figured out why I can't break the limit(only host had the problem 😅), I will patch it once we saw that everything with ghost roles works (also in Dev 2 release).

PR #682, opened 2024-02-21
Retributionist & nemesis

Ultradragon005:
Make retributionist & nemesis ghost-roles. (This means they will only spawn after death)
And some other small things.


Roles/(Ghosts)/Impostor/Nemesis.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                Main.PlayerStates[target.PlayerId].deathReason = PlayerState.DeathReason.Revenge;
+                killer.RpcMurderPlayerV3(target);
+                killer.RpcResetAbilityCooldown();
+                KillCount[killer.PlayerId]--;
+                SendRPC(killer.PlayerId);
+            }
+            else if (Main.AllAlivePlayerControls.Count() < MinimumPlayersAliveToRevenge.GetInt()) killer.Notify(GetString("NemesisTooManyDied"));
+            CustomRoles.Necromancer.IsEnable();
+            return false;
```
Tommy-XL, line 75: What does `CustomRoles.Necromancer.IsEnable();` ?

Ultradragon005: LOL, I forgot to remove it

PR #688, opened 2024-02-22
Create Japanese.txt

SunnyBoiWave:
jp ban words, will add some more later on


Tommy-XL:
BanWords now not supported
But perhaps we will return the settings that ban for prohibited words

SunnyBoiWave:
ahh ok ty, will hold off on spanish one then

PR #702, opened 2024-02-23
Return retri and nemesis

Ultradragon005:
Return is done.
I will now make it compactibility with rolebase (including minion & warden)


Ultradragon005:
image
æugh it dosen't detect the role as a object when I converted, help plz.

Also I for some reason spawned as crewmate base when I did /up sheriff. I was sheriff with crewmate base, it also happened for my alt, but my alt would also sometimes spawn as impostor base like supposed to.

Ultradragon005:
image
these are xml thingy fyi, so it's esier to understand what exact use does the method.

Ultradragon005:
> æugh it dosen't detect the role as a object when I converted, help plz.
Last resort is simply that ghost-roles use old method, I will try for a week to fix this.

Ultradragon005:
actually better idea, I will revert ghost-roles to old method, and do the normal roles instead. and then I will focus on on trying to fix this if possible.

PR #705, opened 2024-02-23
revert nemesis & retributionst

Ultradragon005:
and add bloodmoon & hawk as new roles.

I will do one last check tomorrow to make sure I didn't mess up.

And also I will make it compactible with rolebase once that's completed. So I know what to adjust.


Resources/Lang/en_US.json
```diff
@@ -111,7 +111,8 @@
   "KillingMachine": "Killing Machine",
   "Escapist": "Escapist",
   "Witch": "Witch",
-  "Nemesis": "Nemesis",
+  "Mafia": "Nemesis",
```
Tommy-XL, line 114: It would be better if the role did not have different names in the code and in the string

Tommy-XL: Just name as Nemesis

Ultradragon005: > It would be better if the role did not have different names in the code and in the string

mk, there was just some strings in chatmanager that I was unsure were it pulled that from. But I can try to change it if it's very important. I'm refering to a "after", "Before" thing.

Ultradragon005: okay I found it, will change now.

Ultradragon005: done

Tommy-XL:
Fix conflicts with `RoleBase` pls

Ultradragon005:
done

PR #713, opened 2024-02-24
Crewmate power roles

Ultradragon005:
- [x] Captain
- [x] Copycat
- [x] Dictator
- [x] Guardian
- [x] Lookout
- [x] Marshall
- [x] Mayor
- [x] Monarch
- [x] Overseer
- [x] President
- [x] Swapper
- [x] Telecommunications


Roles/Core/RoleBase.cs
Ultradragon005, line 1: temporary fix untill I fix roles which used another method

Patches/MeetingHudPatch.cs
```diff
@@ -1210,24 +1207,16 @@ public static void Postfix(MeetingHud __instance)
 
                     } */
             //インポスター表示
+            sb.Append(seerRoleClass?.GetMark(seer, target));
```
Tommy-XL, line 1210: Can you sync with the latest commit, since I already added `GetMark` support for modded players

Ultradragon005:
amnesiac will be rewritten later..

Ultradragon005:
I was doing some extra in another branch based from this branch, and found & fixed more problems, so some extra roles:
benefactor, chameleon, coroner

PR #693, opened 2024-02-22
v2.0.0 - Role Base

Tommy-XL:
### Based on: [EndlessHostRoles](https://github.com/Gurge44/EndlessHostRoles)

This is done to make the code shorter and easier

## Estimated completion time in 3 or 4 weeks from 03/03/24
~~It may be completed earlier than planned, in the worst case it will take more than 4 weeks to complete~~
Changes: this will take longer than planned, but this code will be used in `v2.0.0`

### 09/04/24 - all progress done (need test)

Progress: 
- [x] `Impostor ghost`: 100% done (Total 2)
- [x] `Crewmate ghost`: 100% done (Total 2)
- [x] `Impostors`: 100% done (Total 72) (`Evil Mini` done)
- [x] `Crewmates`: 100% done (Total 70) (`Nice Mini` done)
- [x] `Neutrals`: 100% done (Total 69) (Done in: #767)
- [x] Complete Setup for all roles (check this https://github.com/0xDrMoe/TownofHost-Enhanced/pull/693#issuecomment-2002530665)
- [x] Replace `List` with `HashSet` where is possible
- [x] Make `List`, `HashSet`, `Dictionary` as `readonly` and in `Init()` use `Clear()` instead of `= [];` where is possible
- [x] Rework the code execution algorithm in `CheckMurderPatch`
- [x] `RpcMurderPlayerV3` => `RpcMurderPlayer` 
- [x] Add `RpcCheckAndMurder` & `CheckForInvalidMurdering` in some places of the code
- [x] Maybe need making changes to `MurderPlayerPatch` (Done in: #767)
- [x] Sunnyboy is partially broken
- [x] Move execution `OnTaskComplete()` code
- [x] Fix and return ghost roles
- [x] Recode `CheckShapeshiftPatch` 
- [x] Change all `OnShapeshift()`
- [x] Some fix in `KnowRole`

PR #715, opened 2024-02-25
RoleBase : Admirer

NikoCat233:
@Tommy-XL Pls help fix the error in Admirer.cs


Roles/Crewmate/CopyCat.cs
```diff
@@ -322,7 +323,7 @@ public static bool OnCheckMurder(PlayerControl pc, PlayerControl tpc)
                     SabotageMaster.Add(pc.PlayerId);
                     break;
                 case CustomRoles.Admirer:
-                    Admirer.Add(pc.PlayerId);
+                    CustomRoles.Admirer.GetRoleClass()?.Add(pc.PlayerId);
```
Ultradragon005, line 326: yeah so gurge told me this is wrong 😅, anyways ignore copycat I made it rewritten here #713

Patches/PlayerControlPatch.cs
```diff
@@ -2021,7 +2018,7 @@ public static bool Prefix(PlayerControl __instance, [HarmonyArgument(0)] GameDat
                         }
                         else if (tar.Is(CustomRoles.Admirer))
                         {
-                            Admirer.Add(__instance.PlayerId);
+                            CustomRoles.Admirer.GetRoleClass()?.Add(__instance.PlayerId);
```
Ultradragon005, line 2021: put this after the rpcrole has been set.
And get the object from Main.PlayerStates[tar.PlayerId].Role;

Modules/ExtendedPlayerControl.cs
```diff
@@ -1375,7 +1372,7 @@ public static bool KnowSubRoleTarget(PlayerControl seer, PlayerControl target)
             else if (seer.Is(CustomRoles.Egoist) && target.Is(CustomRoles.Egoist) && Egoist.ImpEgoistVisibalToAllies.GetBool())
                 return true;
         }
-        else if (Admirer.IsEnable && Admirer.KnowRole(seer, target)) return true;
+        else if (CustomRoles.Admirer.GetRoleClass().IsEnable && Admirer.KnowRole(seer, target)) return true;
```
Ultradragon005, line 1375: same

Ultradragon005, line 1375: use the native RoleExist() or IsEnable() customroles extension, bcuz this creates a new object.

Gurge44: Yeah this code does not work at all and will always return false since it creates a bew instance which isn't initialized....
Use `Main.PlayerStates.Any(x => x.Value.Role is Admirer {IsEnable: true})` (I typed this out manually so it may have errors)

NikoCat233:
Dont know how to fix these exactly
@Tommy-XL  pls help with this

Tommy-XL:
> Dont know how to fix these exactly @Tommy-XL pls help with this

I don’t understand what errors we are talking about, you can just look at the PR code: #713

NikoCat233:
@Tommy-XL Now i only got this while building
Compiler Error CS1106 Extension methods must be defined in a non generic static class.

Ultradragon005:
> @Tommy-XL Now i only got this while building Compiler Error CS1106 Extension methods must be defined in a non generic static class.

image
had the same thing happen to me when making copycta, internal classes cannot use extension methods for a property (Remove "this" and rewrite the places where the property PlayerControl is extended by using it as a normal method)

NikoCat233:
Uhh

Gurge44:
or move them to a static class (can be internal), like ExtendedPlayerControl.cs

Tommy-XL:
Done https://github.com/0xDrMoe/TownofHost-Enhanced/pull/693/commits/f9fe66d83151ded08022c825174ed0b1133bd691

PR #739, opened 2024-03-11
Fix lawyer

NikoCat233:
If lawyer has no available target, will try to assign other roles for it.
Add a setting where laywer should change role after target dies.


Tommy-XL:
Why not just do it in Role Base, lol

NikoCat233:
LOL

Ultradragon005:
wwoww

PR #736, opened 2024-03-09
RoleBase: Neutral Killing & Evil

Ultradragon005:
- [x] Neutral Killing
- [x] Neutral Evil

PR #747, opened 2024-03-14
Role Base: Neutral Killing & Evil

Tommy-XL:
## Thanks @Ultradragon005 for #736


Roles/Neutral/PlagueBearer.cs
```diff
@@ -71,13 +86,29 @@ public static void SendRPC(PlayerControl player, PlayerControl target)
         writer.Write(target.PlayerId);
         AmongUsClient.Instance.FinishRpcImmediately(writer);
     }
+    public override void ApplyGameOptions(IGameOptions opt, byte playerId)
+    {
+        if(PestilenceList.Contains(playerId))
+            opt.SetVision(PlagueBearer.PestilenceHasImpostorVision.GetBool());
+    }
+    public override bool OnRoleGuess(bool isUI, PlayerControl target, PlayerControl pc, CustomRoles role)
```
Ultradragon005, line 94: I forgot to add a `, ref bool GuesserSuicides `  here (and sync all other OnRoleGuess methods.) instead of this wack get; set;

Tommy-XL: Yeah, that's exactly what I wanted to do

PR #750, opened 2024-03-15
Merge role skill rpc into one

NikoCat233:
Old thing that i forget to do

After role base is finished and released, we can start move all the receive rpc into a override method I think.
The syncroleskill can be used to get the customroles, and pick the rolebase


NikoCat233:
Should be fine to merge

Tommy-XL (approved):

PR #752, opened 2024-03-15
Try fix assign ghost role

Ultradragon005:
It looks like fix was so simple, if this dosen't cause any further issues, check murder can also be replaced.


Ultradragon005:
I just realized I didn't put AmHost in exiled patched 💀 idk how much that mattered, but ig dieplayer also works

PR #753, opened 2024-03-16
Only rpcset ghostroles in dieplayer patch & some fixes

Ultradragon005:
I believe this is more stable hopefully.

### DIFF
--- a/Patches/PlayerControlPatch.cs
+++ b/Patches/PlayerControlPatch.cs
@@ -4421,7 +4421,7 @@ public static class PlayerControlCheckUseZiplinePatch
 [HarmonyPatch(typeof(PlayerControl), nameof(PlayerControl.Die))]
 public static class PlayerControlDiePatch
 {
-    public static void Postfix(PlayerControl __instance, [HarmonyArgument(0)] ref DeathReason Reason, [HarmonyArgument(1)] ref bool assignGhostRole)
+    public static void Postfix(PlayerControl __instance, [HarmonyArgument(0)] ref DeathReason reason, [HarmonyArgument(1)] ref bool assignGhostRole)
     {
         try
         {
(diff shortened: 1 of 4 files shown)


Ultradragon005:
image
not sure what is this used for..

Gurge44:
> image not sure what is this used for..

Literally nothing lmao

Tommy-XL:
This was probably intended to be `PlayerIdList.Contains(killer.PlayerId)` but for some reason it wasn't added, lol

Tommy-XL:
Now it's used? lol
image

Ultradragon005:
image
I just found this, I'll try a prefix to see if that works.

Ultradragon005:
image
uh huh, so it looks to me Like Crew/impGhost are assigned second thing to death. And that GuardianAngel got a special exception to be able to be set even after that. (Because also looking at PlayerControl.DIE() there's a bool assignGhostRole, which runs after crew/impGhost has been assigned...)

So I don't think overriding crew/impghost is required (for checkmurder, idfk know what to do for exiled cuz it runs before exiled), but now everything possible should be pointed towards setting GA, if my asumption is correct.

And this hopefully mitigates BlackScreen errors with Ghost-Roles.

Patches/PlayerControlPatch.cs
```diff
@@ -4436,78 +4421,90 @@ public static bool Prefix([HarmonyArgument(2)] bool fromTop)
 [HarmonyPatch(typeof(PlayerControl), nameof(PlayerControl.Die))]
 public static class PlayerControlDiePatch
 {
-    public static void Postfix(PlayerControl __instance)
+    public static void Postfix(PlayerControl __instance, [HarmonyArgument(0)] ref DeathReason reason, [HarmonyArgument(1)] ref bool assignGhostRole)
```
Ultradragon005, line 4424: added ref DeathReason cuz it might be useful for things like susceptible maybe

PR #756, opened 2024-03-16
fix number chances,  <color=

SunnyBoiWave:


Tommy-XL:
idk, why is this necessary since it already works, lol

PR #758, opened 2024-03-17
Setup Attributes in The Role's Classes

Ultradragon005:
Some Common Attributes,

Also to make some new names because some methods such as "IsCK" may not be so straightforward..


Ultradragon005:
actually nvm

PR #759, opened 2024-03-17
Change lastSequenceId for clients (Hotfix rpc teleport)

Tommy-XL:


Tommy-XL:
It looks like innersloth changed something in how `lastSequenceId` worked on the server and it began to work more stable than before

Tommy-XL:
I tested this code in an online game, and everything seems to work well

PR #765, opened 2024-03-21
Msg for users

Ultradragon005:
Idk why EvilSpirit randomly started working for me without rpcsetrole. so...

I will try that fix ( put ghostrole assign in rpcsetrole prefix) in 2.0.0 dev cuz it might not even be 100% certain that's gonna solve the problem.

I also realized assignghostrole didn't do anything bcuz it was in postfix, but dosen't matter as I should not use that anyways.


Tommy-XL:
I think I will temporarily use `ResetCam();` for ghost roles

PR #768, opened 2024-03-22
Increase character limit for sending chat messages

Tommy-XL:
There is a possibility that the message may be too large to send and he cannot send it to the player


Tommy-XL:
Seems work fine

PR #777, opened 2024-03-26
Optional proxy to avoid issues while logging in

Pietrodjaowjao:
dont merge yet, i will wait for moe to send the proxy address first

### DIFF
--- a/main.cs
+++ b/main.cs
@@ -111,6 +111,7 @@ public class Main : BasePlugin
     public static ConfigEntry<string> WebhookURL { get; private set; }
     public static ConfigEntry<string> BetaBuildURL { get; private set; }
     public static ConfigEntry<float> LastKillCooldown { get; private set; }
+    public static ConfigEntry<bool> UseProxy { get; private set; }
     public static ConfigEntry<float> LastShapeshifterCooldown { get; private set; }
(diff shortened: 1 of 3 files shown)


main.cs
```diff
@@ -448,12 +449,22 @@ public override void Load()
         GodMode = Config.Bind("Client Options", "GodMode", false);
         AutoRehost = Config.Bind("Client Options", "AutoRehost", false);
 
+        //Proxy
+        UseProxy = Config.Bind("Client Options", "UseProxy", false);
```
Tommy-XL, line 453: Client options also need add in `ClientOptionsPatch`

Pietrodjaowjao: ik, but right now im using only for the .cfg file

Pietrodjaowjao: later i will add it

Pietrodjaowjao:
chat is this correct

Pietrodjaowjao:
it works but this proxy does not exist 💀

Pietrodjaowjao:
so like

Pietrodjaowjao:
image

Pietrodjaowjao:
but when the proxy exists, it works just fine 😎 
image

Pietrodjaowjao:
i might have fixed the issue with another fix.

PR #780, opened 2024-03-27
Start ReceiveRPC Override + Improvements

Ultradragon005:
did all the ones in syncroleskillreader as it was simple, and @NikoCat233 may do the rest.
Basically ready to merge, I'm just adding/changing stuff as I see fit rn

Did these points for all roles now.
- [x] Standarized setup for all roles
- [x]  Replace List with HashSet where is possible
- [x]  Make List, HashSet, Dictionary as readonly and in Init() use Clear() instead of = []; where is possible
- [x] RpcMurderPlayerV3 => RpcMurderPlayer
- [x] Add RpcCheckAndMurder & CheckForInvalidMurdering in some places of the code

### DIFF
--- a/Modules/ExtendedPlayerControl.cs
+++ b/Modules/ExtendedPlayerControl.cs
@@ -640,17 +640,6 @@ static class ExtendedPlayerControl
     }
     public static void RpcMurderPlayerV3(this PlayerControl killer, PlayerControl target)
     {
-        if (target.Is(CustomRoles.Pestilence) && killer != target) { killer.SetRealKiller(target); (killer, target) = (target, killer); Main.PlayerStates[target.PlayerId].deathReason = PlayerState.DeathReason.PissedOff; goto postPest; }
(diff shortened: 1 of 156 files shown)


Ultradragon005:
Standard setup should be done for all roles nowm I think I already did all neutrals.

Modules/ExtendedPlayerControl.cs
```diff
@@ -640,40 +640,18 @@ public static void RpcExileV2(this PlayerControl player)
...
-
-        if (target.Is(CustomRoles.Susceptible))
-            Susceptible.CallEnabledAndChange(target);
-
-        if (killer.PlayerId == target.PlayerId && killer.shapeshifting)
+        if (killer.PlayerId == target.PlayerId && killer.shapeshifting) 
         {
             _ = new LateTask(() => { killer.RpcMurderPlayer(target, true); }, 1.5f, "Shapeshifting Suicide Delay");
             return;
         }
```
Ultradragon005, line 647: was this something important? Should jsut simply be moved to rpcmurder then.

Tommy-XL: No, it's not needed
image

Ultradragon005:
Ig this is finished

PR #784, opened 2024-03-30
Return Horse mode and prepare for 1st april update

Pietrodjaowjao:


Pietrodjaowjao:
Finally fixed all the issues bro, i recommend to merge this in 1st april

Pietrodjaowjao:
ig you could just merge this... there is 2 days remaining for the release lol and the update was already leaked

Ultradragon005:
I think people are celebrating easter

Pietrodjaowjao:
just like 3 hours remaining

Tommy-XL:
I was planning to add this for v2.0.0

Pietrodjaowjao:
why?

Tommy-XL:
Because the next version 1.6.0 is planned for full release, but without ghost roles, especially since the April Fool's joke will go on for another week and may cause some problems

Tommy-XL:
image

Wouldn't it be better to always display this banner?
Since if the host activates it, then it will be activated for others clients

And Horse Mode will remain a client setting

Pietrodjaowjao:
cant we make the gamemode change midgame instead

Pietrodjaowjao:
when toggle is on

Tommy-XL:
But it is activated only for host

Pietrodjaowjao:
i mean that we can change GameMode then sync rpc, so body types will change

Tommy-XL:
This is not bad, but the current code doesn't do it?

Pietrodjaowjao:
Yeah it doenst, i will fix tomorrow.

Pietrodjaowjao:
ig its impossible bc other players would need to rejoin

Pietrodjaowjao:
just merge after the april 8 maybe, the event will end that time

Tommy-XL:
> image
> 
> Wouldn't it be better to always display this banner? Since if the host activates it, then it will be activated for others clients
> 
> And Horse Mode will remain a client setting

Well, you can just do it as I suggested, the current code only applies to modded players, although it can be done for other players, although you will have to create new lobbies for this

Pietrodjaowjao:
ok

Pietrodjaowjao:
There is a way to fix the heads clipping trought walls, its just a shader issue, but since it will only work for who have tohe installed, ig i will not put

Pietrodjaowjao:
Ready to merge

PR #787, opened 2024-03-31
Automatically change the RoleName In LongInfo

Ultradragon005:


Ultradragon005:
I realized it will not work for other languges 💀

PR #789, opened 2024-04-01
 Automate Lang.dat And give roleinfo a new look (+ Colors)

Ultradragon005:
`1.` "Role:" will now be automatically replacing InfoLong if it is custom. (Works for all languages now :D)

`2.`Split /m Into: 1.  Role & desc, 2. Role Config, 3. Addons & desc (html tags removed if it's too long). (I mean for all instances which do that e.g firstmeeting, /r role etc..)

/r role is just 1. role & desc, 2. role config ofc. Ghostmsg only shows the role.

I figured since rolebase has the core done I'd do this now.

`3.` Added a new <title> </title> tag which let's you adjust the tempalte title.


Ultradragon005:
Erm... so for somereason non-english languages have to do t +f5 after entering lobby to grab the name. I'll see if I can find the problem of this, or I just forceload if for non-english 😂

Ultradragon005:
> Erm... so for somereason non-english languages have to do t +f5 after entering lobby to grab the name. I'll see if I can find the problem of this, or I just forceload if for non-english 😂

fixed, I forgot to implicintly take language.

Ultradragon005:
should be done now

PR #805, opened 2024-04-09
Make dbconnect an ienumerator to avoid lag and even connection issues

Pietrodjaowjao:


Pietrodjaowjao:
Will do for updater very soon too

Pietrodjaowjao:
I guess scene change will not affect the dbconnect, needs testing

PR #807, opened 2024-04-10
Some Changes In Bloodmoon

Ultradragon005:


Roles/(Ghosts)/Impostor/Bloodmoon.cs
```diff
@@ -94,12 +93,13 @@ public override void OnOtherTargetsReducedToAtoms(PlayerControl target)
...
-    private string OthersNameText(PlayerControl seer, PlayerControl player, bool IsForMeeting) 
+    private string OthersNameText(PlayerControl seer, PlayerControl player, bool IsForMeeting, bool isforhud = false) 
     {
-        if (IsForMeeting || seer != player) return string.Empty;
+        var IsMeeting = GameStates.IsMeeting;
+        if (IsMeeting || seer != player || player == null || isforhud) return string.Empty;
         var playerid = player.PlayerId;  
 
-        if (LastTime.TryGetValue(playerid, out var lastTime) && lastTime + 1 <= GetTimeStamp() && !IsForMeeting)
+        if (LastTime.TryGetValue(playerid, out var lastTime) && lastTime + 1 <= GetTimeStamp() && !IsMeeting)
```
Tommy-XL, line 102: text in `isForHud` will not be displayed for modded players

And `GameStates.IsMeeting` doesn't work correctly in NotifyRoles, so better use `IsForMeeting`

Ultradragon005: When I used IsForMeeting, the target got killed during meeting..

I think it's maybe bcuz it returns false here? 
image

Tommy-XL: Then use `isForMeeting || GameStates.IsMeeting` and return `string.Empty`

Ultradragon005: done

PR #826, opened 2024-04-13
Setup Simple Attributes (Done right)

Ultradragon005:
I could also automate setupcustom options via:
image


Ultradragon005:
Oh I realized alot of customroleoptions have been removed, so I think I'm gonna change the approach

Ultradragon005:
gonna use the latest commit instead

PR #850, opened 2024-04-26
Fixes and some changes In "HudSpritePatch"

Tommy-XL:
Update custom sprites only when necessary, as it updates sprites every second

Thanks @D1GQ for fixing a bug where custom sprites are not sets correctly and they become too bright or, conversely, too dark

### DIFF
--- a/Patches/HudPatch.cs
+++ b/Patches/HudPatch.cs
@@ -209,15 +209,17 @@ class SetVentOutlinePatch
 class SetHudActivePatch
 {
     public static bool IsActive = false;
-    public static void Postfix(HudManager __instance, [HarmonyArgument(2)] bool isActive)
+    public static void Postfix(HudManager __instance, [HarmonyArgument(0)] PlayerControl player, [HarmonyArgument(2)] bool isActive)
     {
+        // Fix vanilla bug when report button displayed in the lobby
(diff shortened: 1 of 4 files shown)


Tommy-XL:
> @Tommy-XL There seems to be a bug for shapeshifter roles, where when the shapeshift duration is over if there is one, the icon gets set back to default.

It's not bug, some roles that change the shapeshift status remove the custom icon if necessary

image

D1GQ:
> It's not bug, some roles that change the shapeshift status remove the custom icon if necessary
> 
> image

It was happening to Doll Master and there's no conditions set for it! and this issue has only occurred after I added your fixes.
Screenshot 2024-04-26 073628
So it might be something on my side or it could be a bug.

Tommy-XL:
Try this code

image

```
        // Set custom icon for ability button (Shapeshift, Vitals, Engineer Vent)
        __instance.AbilityButton.graphic.sprite = newAbilityButton;
        __instance.AbilityButton.usesRemainingSprite.sprite = newAbilityButton; // when Shapeshift duraction is remaining
```

Tommy-XL:
Hmm, I reproduced the bug, it's not fixed

Tommy-XL:
Fixed

PR #851, opened 2024-04-26
Make killog and lastroles colored again !! + fix blackscreen (hopefully) + missing strings+ fix messages not sent vanilla player

Ultradragon005:
This makes it so code split the text if explicitly told so and text is too long, other-wise remove htmltags.


Gurge44:
You're welcome

PR #852, opened 2024-04-27
Remove limit of rpc call numbers

Pietrodjaowjao:

PR #856, opened 2024-05-01
Fix entire game broken

Ultradragon005:
reverts https://github.com/0xDrMoe/TownofHost-Enhanced/pull/852/ , because all current rpcs are based on byte it does not recognize them I think.

Perhaps you could make a new enum tho "Extendeed_RPC".

### DIFF
--- a/Helpers/WriterHelper.cs
+++ b/Helpers/WriterHelper.cs
@@ -2,7 +2,7 @@
 
 namespace TOHE;
 
-public static class WriterHelper
+public static class WriterHelper // Currently not in use, because all rpc are based on byte. May be in use for a new enum perhaps...
 {
     /// <summary>
     /// Starts a rpc using a integer value.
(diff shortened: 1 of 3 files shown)


Ultradragon005:
gonna make a seperate pr to fix the colors cutting off cuz it will take sometime

Pietrodjaowjao:
dont need to revert the entire pr

Ultradragon005:
but, there's a way to make smt like `public virtual void RecieveExtendedRpc(out Dictionary<string, Action<MessageReader, Playercontrol, bool, bool>)`; if we rlly need to

Pietrodjaowjao:
Bro changing WriterHelper to private will not change anything 💀 , the rpcs are already using byte, only rpcs that use int will use WriterHelper thing

Ultradragon005:
mk, it had 0 references so I just made it that untill it is gonna be used.

PR #858, opened 2024-05-01
Fix color cutting

Ultradragon005:
Merge it after https://github.com/0xDrMoe/TownofHost-Enhanced/pull/856/ , because I built ontop of it to not get broken messages.


Ultradragon005:
if a HUGE influx of messages is being sent it may send them in the incorrect order being of ping prioritizing some over others, so you may see very minor odities sometimes (such as just a random x1 when doing /r). But other than that works well

PR #873, opened 2024-05-05
Full Reset Cam Player

Tommy-XL:
Added a new solution for anti black screen

After meeting, players select a dead player and local teleport him off the map and kill him there and return to the starting point again

Possible problems:
Too many RPCs sent can cause a disconnects
The player/ghost may not be teleport due to problems with the servers


NikoCat233:
Dont even need to teleport the dead player or killer
just use rpc specific murder and ignore the report body from the killer

We should also add a new bool in playerstates, which decides whether the body is reportable (ppls dead last round become unreportable after the meeting)

Ultradragon005:
image
lol

NikoCat233:
Thats not the case

ryuk201198:
Instead of teleporting and killing we can use `RpcExileV2()`? Hangman used this for thr kills and doesn't leave a deadbody behind

NikoCat233:
thats completely not the case
We need the killer to kill someone to remove the blackout of the killer

Ultradragon005:
> Instead of teleporting and killing we can use `RpcExileV2()`? Hangman used this for thr kills and doesn't leave a deadbody behind

it dosen't do anything with the killer it just does target.Exiled().

Tommy-XL:
> Dont even need to teleport the dead player or killer just use rpc specific murder, send the murderplayer rpc only to the killer and ignore the report body from the killer
> 
> We should also add a new bool in playerstates, which decides whether the body is reportable (ppls dead last round become unreportable when a meeting starts)

This will cause confusion among the players, I teleport ghost outside the map so that he is not visible on the map (but will still be visible in meetings as killed)

NikoCat233:
Just use specific murder player so its only visiable to killer

Tommy-XL:
> Just use specific murder player so its only visiable to killer

The code already does this, but all unmodded players do it

NikoCat233:
Ok no problem

Pietrodjaowjao:
yeah, it will always teleport if you use rpcteleport

Tommy-XL:
I need to do something about this
And I don’t know whether it will be implemented in the end or not.

Ultradragon005:
but wait. idk if I was tripping or not.

But, I was blackscreened in a match. But after a natural reactor happened shortly after meeting. I was no longer blackscreened and could also move in that same match.
But I don't know how to replicate a blackscreen manually expect for doing it in a local match with my phone but that isn't very accurate to how it actually behaves.

Tommy-XL:
This is no longer necessary as changing the base role fixes a lot of black screens

PR #884, opened 2024-05-09
Re-factor codebase roles + fix many issues

Ultradragon005:
In this pr:
- Automate "IsEnable".
- Automate syncskillrolerpc. you will litteraly just need to use "SendSkillRpc()" where neccessary.
     - Make an non-static AbilityLimit count to use for roles with usage limits.
- Remove the need for doing playeridlist, instead use a common object method to fetch the player. _(where playeridlist is used it should be removed if it is inside a object method as it makes it run twice, but not too important rn.)_
     
- Remove experimental tab, and instead make it so if any roles are experimental it will display at the top of the tab (with a unique segment identifier jsut like the others), they're there to be tested so we can resolve the issues

- Made some roles compatible with multiple players by using object attributes (even if a role is maximum 1 person, some roles like amnesiac/imitator/copycat are able to copy roles and this would allow to copy more roles) (maybe I'll do all singleroleoptions one, but another day..)

- Made it so that ALL roles use syncskillrolereader by using their object class.

Also fixed the killer ghastly chose to not have a dead target


Ultradragon005:
btw, was making "hater", "Taskinator", "Collector", "Cursed Soul", isenable to false intentional?

Roles/Core/RoleBase.cs
```diff
@@ -1,33 +1,57 @@
...
+    }
+
+    public void OnAdd(byte playerid) // The player with the class executes this
+    {
+        _state = Main.PlayerStates.Values.FirstOrDefault(state => state.PlayerId == playerid);
+        try { 
+            CustomRoleManager.RoleClass.FirstOrDefault(r => r.Key == _state.MainRole).Value.IsEnable = true;
+            this.IsEnable = true; // Not supposed to be used, but some methods may have still implemented that check.
+        } catch { } // temporary try catch
+
```
Ultradragon005, line 32: okay it's ready, but will have to change this once all the current role pr's are merged (it's temporarily in a try catch bcuz of that)

Ultradragon005:
and the amount of files changed is basically just because I did use quick remove to remove all isenable overrides cuz it wasn't needed anymore.

Ultradragon005:
goddamn, finnaly fixed it

Ultradragon005:
Some things I've discovered during this:
- Unlike `PlayerControl.AllPlayerControls`, `playerstate` does not clear when player left the game (which actually very good for this case as many roles already used non-static attributes and had to activate by the player and not just the roleclass so I've changed allenabledroles accordingly)

- if `any(x => bool)` is followed by another linq method to filter out (such as where), it will still try to run the any method but always return false on it. which makes it important to check first if `.any()` exist in the ienumerable else return null.

Tommy-XL:
It seems to me that not all roles correctly send RPCs, `RpcInx` should always go first and then everything else, but somewhere either there is a package for `RpcInx` or it is not there at all, and somewhere it is sent almost second or last in order

Screenshot_20240517_084124
Screenshot_20240517_084058

Ultradragon005:
Should be fixed now. and also rpcindx isn't obligatory but I made sure it dosen't send any errors

Ultradragon005:
image
image

I tried to test recieverpc as modded host but ig that dosen't work :/  but if it was working previously it will work now. 

(I tested judgerpc on 1.6.0 just to make sure it wasn't a error in my branch)

Modules/RPC.cs
```diff
@@ -877,35 +645,248 @@ public static void SetCustomRole(byte targetId, CustomRoles role)
...
             {
-                //Crew Roles
-                case CustomRoles.Admirer:
-                    Admirer.ReceiveRPC(reader, false);
+                case Admirer ad when RpcInx == 1:
+                    ad.ReceiveRPC(reader, false);
+                    break;
+                case Admirer ad when RpcInx == 2:
+                    ad.ReceiveRPC(reader, true);
+                    break;
```
Tommy-XL, line 675: I think this method with `RpcInx` is much better (since there is no need to add RPC manually)
But this is not necessary to change it

image

Tommy-XL:
And fix conflicts

Ultradragon005:
don't merge yet, have to change some stuff cuz of recent merges.

Ultradragon005:
it is ready again

Modules/RPC.cs
Ultradragon005: oh I see, well we can do it for every new role. And if I have time to make a new pr to make all of the roles like that..

PR #885, opened 2024-05-09
New Imp Role: Dollmaster

D1GQ:
# Roles:
## Dollmaster (Imposter)
DollMaster
As the Dollmaster, you can temporarily take control of any player by using the Shapeshift button and to make them do your Deeds, be careful though while possessing, the person you possess and your main body is fragile to killing roles!


D1GQ:
@ryuk2098 Here is the first role!

Roles/Impostor/DollMaster.cs
```diff
@@ -0,0 +1,335 @@
...
+
+    public override void SetKillCooldown(byte id) => Main.AllPlayerKillCooldown[id] = DefaultKillCooldown.GetFloat();
+
+    // A quick check if a player is being possessed.
+    public static bool IsDoll(byte id)
+    {
+        if (ReducedVisionPlayers.Contains(id))
+            return true;
+        return false;
+    }
```
Tommy-XL, line 78: Here you can simplify the code by replacing it with:
`public static bool IsDoll(byte id) => ReducedVisionPlayers.Contains(id);`

Roles/Impostor/DollMaster.cs
```diff
@@ -0,0 +1,335 @@
...
+    {
+        if (!IsControllingPlayer) return string.Empty;
+        if (GameStates.IsMeeting) return string.Empty;
+        if (target == null) return string.Empty;
+        if (controllingTarget == null) return string.Empty;
+        if (DollMasterTarget == null) return string.Empty;
+        if (seer.PlayerId != target.PlayerId && target.PlayerId == controllingTarget.PlayerId) return "<color=#ffea00>" + GetString("DollMaster_MainBody");
+        if (seer.PlayerId == target.PlayerId && target.PlayerId == DollMasterTarget.PlayerId) return "<color=#ffea00>" + GetString("DollMaster_Doll");
+        return string.Empty;
+    }
```
Ultradragon005, line 330: better to use GetLowerTextand also in `"<color=#ffea00>" + GetString("DollMaster_MainBody");` and `"<color=#ffea00>" + GetString("DollMaster_Doll");`, remember to end it with a `+ "</color>"`

Roles/Impostor/DollMaster.cs
```diff
@@ -0,0 +1,335 @@
...
+
+    // Prepare for a meeting if possessing.
+    public override void OnReportDeadBody(PlayerControl pc, PlayerControl target) // Fix crap when meeting gets called.
+    {
+        if (IsControllingPlayer && controllingTarget != null && DollMasterTarget != null)
+        {
+            bool shouldAnimate = false;
+            Utils.GetPlayerById(DollMasterTarget.PlayerId).RpcShapeshift(Utils.GetPlayerById(DollMasterTarget.PlayerId), shouldAnimate);
+            Utils.GetPlayerById(controllingTarget.PlayerId).RpcShapeshift(Utils.GetPlayerById(controllingTarget.PlayerId), shouldAnimate);
+            UnPossess(Utils.GetPlayerById(DollMasterTarget.PlayerId), Utils.GetPlayerById(controllingTarget.PlayerId));
```
Tommy-XL, line 178: I don't think you need to use `GetPlayerById` because they already contain `PlayerControl`

image

Tommy-XL: There are similar cases in many parts of the code

Roles/Impostor/DollMaster.cs
Ultradragon005: image
same here

Roles/Impostor/DollMaster.cs
D1GQ: Yeah sorry about that this is one of my first roles at the time and I was using a different method to get the player IDs back then, so I'm going to do a general cleanup right now.

Tommy-XL:
Fix conflicts pls

D1GQ:
> Fix conflicts pls

Fixed!

PR #893, opened 2024-05-12
Automate Fortune Teller Show-roles

Ultradragon005:
It shows 6 random roles which contains target's role. (it could be made to set values, but I think this is neat)


0xDrMoe:
Is this ready?

Ultradragon005:
> Is this ready?

yeah

main.cs
```diff
@@ -676,7 +676,6 @@ public enum CustomRoles
     Oracle,
     Overseer, 
     Pacifist, 
-    Paranoia, //paranoid
```
Tommy-XL, line 679: You forgot remove role from `roleColor.json` and from english strings

PR #896, opened 2024-05-14
Spawn timeout manager

NikoCat233:
Add a new manager to check whether client is ready instead of adding a latetask that usually wont work

Check by:
client send check color rpc
client custom nt is sending data

Can do
Better autohost kick
Better MessageToSend Remove

To do later
Completely rework the current green bean delay into Spawn timeout manager

Move rpcversion check to spawn timeout manager. It makes sense right?


NikoCat233:
Most time green bean is caused by client side network issue (sometime host lag or server lag)
the client cannot spawn it self correctly so it wont send rpc and data, result in green beans.
It doesnt have much relevance to the color, name or other stuffs

Tommy-XL:
Can you sync pr with RoleBase?

NikoCat233:
Have no idea what i should change

PR #908, opened 2024-05-18
Remove custom nt check

NikoCat233:
InnerSloth is not presenting CustomNT data as how its originally sent by client

In real testing even when custom nt is despawned client side, innersloth will try send a data per 1-2 second


NikoCat233:
3 seconds is ok in this case

PR #909, opened 2024-05-18
Remove splitmessage from lastroles

Ultradragon005:
So after messing around I accidentaly discovered that the title has a limit of like 2400~~ chars and lastroles is never more than like 1370 so we gud af.


Ultradragon005:
image
nvm my dumbass forgot to realize the loggers didn't have any colors in them

PR #914, opened 2024-05-20
Revert: Re-factor codebase roles

Tommy-XL:


Tommy-XL:
Veeeeeeery buggy

Tommy-XL:
And the improvement for Chronomancer is also reverted
It is necessary to return the improvement later to the current role base

PR #915, opened 2024-05-20
Revert rpc

Ultradragon005:


Ultradragon005:
It's a simple fix, but to make it easy. I am going to revert it, and then make a new pr which has the change (but fixed version.)

Ultradragon005:
There's quite a bit of roles and text fixes I did which would also get reverted with reverting codebase roles

Ultradragon005:
almsot done

Ultradragon005:
should be done now. Just need a review from niko, since I write the playerid as intpacked instead now (so that it's possible to access the playerclass)

Roles/Neutral/Follower.cs
```diff
@@ -60,10 +60,10 @@ public override void Add(byte playerId)
         if (!Main.ResetCamPlayerList.Contains(playerId))
             Main.ResetCamPlayerList.Add(playerId);
     }
-    private static void SendRPC(byte playerId)
+    private void SendRPC(byte playerId)
     {
         MessageWriter writer = AmongUsClient.Instance.StartRpcImmediately(PlayerControl.LocalPlayer.NetId, (byte)CustomRPC.SyncRoleSkill, SendOption.Reliable, -1);
-        writer.WritePacked((int)CustomRoles.Follower); //SyncFollowerTargetAndTimes
+        writer.WritePacked(_state.PlayerId); //SyncFollowerTargetAndTimes
         writer.Write(playerId);
```
Tommy-XL, line 67: Maybe instead of `_state.PlayerId` shold `_Player`

Ultradragon005: lol that was from a regex find and edit. but I realized follower rpc isn't used. Ig I can activate it (idk if works properly tho since it wasn't previously used)

Roles/Impostor/BountyHunter.cs
```diff
@@ -74,16 +74,18 @@ public override void Remove(byte playerId)
     private static void SendRPC(byte bountyId, byte targetId)
     {
         MessageWriter writer = AmongUsClient.Instance.StartRpcImmediately(PlayerControl.LocalPlayer.NetId, (byte)CustomRPC.SyncRoleSkill, SendOption.Reliable, -1);
+        writer.Write(bountyId);
```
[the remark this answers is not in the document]
Ultradragon005: forgot to make it use setbountytargetrpc, did it now.

PR #917, opened 2024-05-20
Automatically ban spoofed fcs

Pietrodjaowjao:
Instead of only checking for empty friend codes, check for any kind of spoof.


Pietrodjaowjao:
Shit

Tommy-XL:
I think that here we need to change the context of the reason for the player’s kick

image

Pietrodjaowjao:
Okk

PR #918, opened 2024-05-21
Cancel Vote System

Ultradragon005:
The first vote is for skill and will be returned.
The second vote will be your real vote.

Role to work with

- [x]  Cleanser
- [x]  Eraser
- [x]  Fortune teller
- [x]  Tracker
- [x]  Keeper 
- [x]  Oracle
- [x]  Godfather
- [x] Soul Collector


NikoCat233:
Dont merge yet, its buggy

NikoCat233:
You should add a option setting to allow /vote commands
and make /vote 253 as skip vote (just do castvote 253)

also a setting to spam chats on received /vote

NikoCat233:
Ok it seems good

NikoCat233:
image
Can u do this? I think it makes more sense

Tommy-XL:
I think merge this after dev 17 or until there is a more or less stable dev version

Tommy-XL:
And also, some roles have a “hide vote” setting, I think now it doesn't make any sense

Ultradragon005:
> And also, some roles have a “hide vote” setting, I think now it doesn't make any sense

oh right. gonna just remove that.

Patches/MeetingHudPatch.cs
```diff
@@ -190,8 +190,6 @@ public static bool Prefix(MeetingHud __instance)
                 var player = Utils.GetPlayerById(ps.TargetPlayerId);
                 var playerRoleClass = player.GetRoleClass();
 
-                // Hide roles vote
-                if (playerRoleClass.HideVote(ps)) continue;
```
Tommy-XL, line 194: I think need leave this feature as some new roles may use this

Or just comment out this code so as not to lose it

Ultradragon005: hmm okay

Tommy-XL:
I think this system can be improved, for example if the player does not want to use the ability he can click on skip voting and go to the real voting

Tommy-XL:
Can you fix conflict?

Ultradragon005:
fixed

Ultradragon005:
image

PR #919, opened 2024-05-22
New System: Enhanced File Checker

D1GQ:
# Overview
Added a new system called the ```Enhanced File Checker```, this checks the game files of Among Us and checks what BepInEx mods are running in the process, if it detects any unauthorized files or mods it will deny access to online play, and later on it is planned to send information to the API of the of the players Name/FriendCode/PUID and tags of what got detected. This should add an extra layer of protection people trying to cheat while using the mod!
Is_a_cheater
## Futures:
```diff
+ | New and improved EAC List check.
+ | Disable play button until EAC list it's gathered.
+ | BepInEx mod ban list.
+ | Among Us version text word checker / ban list.
+ | Sicko, Among Us Menu, and Malum Menu detection.
+ | Player information logger.
! | Fixed EAC bypass.
! | And much more!
- | This system will automatically be disabled for dev's for testing the anti-cheat.
```


Ultradragon005:
couldn't they just build the source code without it tho..

D1GQ:
> couldn't they just build the source code without it tho..

The average Joe wouldn't really know how to do that though.

Pietrodjaowjao:
why moe keeps commiting here

Ultradragon005:
Nah, I think it's because merging, makes the commit appear on the branch

D1GQ:
> Nah, I think it's because merging, makes the commits appear on the branch

That is correct

Patches/CredentialsPatch.cs
```diff
@@ -214,6 +214,7 @@ public static void Prefix(ModManager __instance)
 
         LateTask.Update(Time.deltaTime);
         CheckMurderPatch.Update();
+        EFC.UpdateUnauthorizedFiles();
```
Tommy-XL, line 217: This code will be constantly checked during the game, won't it?

Tommy-XL: I mean I don't want it to affect performance when the mod is constantly reproducing this code

D1GQ: All the code is doing is checking if it did find an unauthorized file, checking for the unauthorized files only gets run when someone presses on the play button, so I don't think it should decrease the performance, but if it's a big issue I can move it back to the update on the main menu

Pietrodjaowjao: theres no reason to do this, mods cant be added in runtime lol

Pietrodjaowjao: Please patch MMOnlineManager.Start or something like this

Tommy-XL: Yeah, this will probably be better than a constant check that is checked 20-30 times in one second

D1GQ: I'll do something to take down the stress

0xDrMoe:
This can be closed due to BAU now, correct?

D1GQ:
Yeah

PR #926, opened 2024-05-24
Missing Dollmaster Strings

Ultradragon005:


Patches/ChatCommandPatch.cs
```diff
@@ -834,7 +834,7 @@ public static bool Prefix(ChatController __instance)
                     if (GameStates.IsHideNSeek) break;
                     if (!(DebugModeManager.AmDebugger && GameStates.IsInGame)) break;
                     if (GameStates.IsOnlineGame && !PlayerControl.LocalPlayer.FriendCode.GetDevUser().DeBug) break;
-                    subArgs = text.Remove(0, 3); 
+                    subArgs = text.Remove(0, 11);
```
Ultradragon005, line 837: I did an oopise, meaning this is supposed to be 11

PR #939, opened 2024-05-26
Add-Lobbytimer-in-template

NikoCat233:
image


Ultradragon005:
😭 this is crazy

NikoCat233:
wtf why is this getting rejected

GD825:
my host came up with a suggestion.. she created a command for it as well /t clock, which will show them the timer when the player asks for it.. i think that would be a great add to that, so this way u can kill off the big on screen one, and the player can just use the command to see the game timer

PR #941, opened 2024-05-27
Some improvements for eac

NikoCat233:
1. Players can report hack murdered body in lobbies, only report and wont ban
2. Add some basic gamedata flags check


Tommy-XL:
There are fears that this will lead to a huge number of bugs, so I'll probably merge this to dev 24

Pietrodjaowjao:
It wont cause bugs

PR #942, opened 2024-05-28
Some Bloodmoon Change

Ultradragon005:


Ultradragon005:
ported colelction extensiosn from EHR

Tommy-XL:
Lol

image

Ultradragon005:
bruh what the heck

Modules/Extensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+            int n = list.Count;
+            var r = IRandom.Instance;
+            while (n > 1)
+            {
+                n--;
+                int k = r.Next(n + 1);
+                (list[n], list[k]) = (list[k], list[n]);
+            }
+            return list;
+        }
```
Tommy-XL, line 41: This already exists in the code

image

Ultradragon005:
fixed github tripping ballz

Gurge44:
I would've liked to see a link to my original file at the top, guess they don't do that here

Ultradragon005:
oh okay, gonna ask tommy to add this so I don't need to make another pr

Tommy-XL:
> I would've liked to see a link to my original file at the top, guess they don't do that here

Ok, i will added it

Tommy-XL:
> I would've liked to see a link to my original file at the top, guess they don't do that here

https://github.com/0xDrMoe/TownofHost-Enhanced/commit/08b1dcf0005bcc53f32710041d41f10eceab8b8e

Ultradragon005:
> Also worth mentioning that this `Do` method will make the iteration run in parallel between the elements, so the order of completion may be different from the order of elements in the collection, but it makes a huge performance improvement (and in 99% of cases we don't care about the order of completion) But this may not work well with BepInEx (like async)

Multi-Threading at it's finest 😼 but unfortunately it can lead to crash sometimes

Ultradragon005:
but ig the other parrarel methods run just fine, I noticed there was no lag when opening the settings manager now.

Gurge44:
> but ig the other parrarel methods run just fine, I noticed there was no lag when opening the settings manager now.

that was because of another thing ported from EHR lol

Gurge44:
> > Also worth mentioning that this `Do` method will make the iteration run in parallel between the elements, so the order of completion may be different from the order of elements in the collection, but it makes a huge performance improvement (and in 99% of cases we don't care about the order of completion) But this may not work well with BepInEx (like async)
> 
> Multi-Threading at it's finest 😼 but unfortunately it can lead to crash sometimes

BepInEx isn't really friends with async or parallel lol

Ultradragon005:
> > but ig the other parrarel methods run just fine, I noticed there was no lag when opening the settings manager now.
> 
> that was because of another thing ported from EHR lol

oh.

Gurge44:
Basically everything in System.Threading is useless

Gurge44:
So I think you should remove the other Parallel stuff because it will crash the game too

Ultradragon005:
ye I realize, gonna do it in my lates pr

Tommy-XL:
Game crash due `CustomRpcSender`  so I temporarily removed the code

image

PR #943, opened 2024-05-28
Fix Duplicates in OtherIcons

Ultradragon005:
static methods are already filtered out bcuz it's a hashset.

I think this should work gonna test it rn.


Ultradragon005:
image
looks good!

The reason 15 files changes is because I merge latest to be able and put it in collectionextensions

Tommy-XL:
> The reason 15 files changes is because I merge latest to be able and put it in collectionextensions

Wait, sync not helped?

Ultradragon005:
> Wait, sync not helped?

well, sync is what caused this lmao, but I will open a new pr, which will fix it

PR #944, opened 2024-05-28
Fixes black screens during role assign and fix Scientist assign for desync roles

Tommy-XL:
Try to prevent players from getting a black screen when any player quits the game while assign roles


Tommy-XL:
I’m not sure exactly if this will help, but I’m surprised that there are no null checks for players anywhere before sending RPCs

Tommy-XL:
`SetRole` activates the `CoShowIntro`, so if the player has a black screen, it means that the player did not receive the RPC about getting the role
image

Tommy-XL:
Unfortunately, I can't really test this, but I hope it’s work fine

PR #947, opened 2024-05-28
Add ovverides (MarkOther, SuffixOther, LowerOther)

Ultradragon005:
New ovverrides. uhhh.... I may or may not decided to do this just to make filtering instead of just marking static make sense 😎


Ultradragon005:
oh god... so many unity.grpahic using

Ultradragon005:
remove the parrarel things because it's bepinex biggest enemy or smt

Roles/Core/CustomRoleManager.cs
```diff
@@ -414,7 +410,8 @@ public static string GetMarkOthers(PlayerControl seer, PlayerControl seen, bool
     /// </summary>
     public static string GetLowerTextOthers(PlayerControl seer, PlayerControl seen, bool isForMeeting = false, bool isForHud = false)
     {
-        if (!LowerOthers.Any()) return string.Empty;
+        var LowerOthers = new HashSet<Func<PlayerControl, PlayerControl, bool, bool, string>>();
+        LowerOthers = AllEnabledRoles.Select(lower => (Func<PlayerControl, PlayerControl, bool, bool, string>)lower.GetLowerTextOthers).FilterDuplicates();
```
Tommy-XL, line 414: idk, maybe the old method is better, because this code will check the list every second for duplicates and Idk if this will affect performance

Ultradragon005: alright I'll make it added on start

Tommy-XL (approved):

PR #952, opened 2024-05-31
Role basis desync and remove disable votekick option

Pietrodjaowjao:
self explanatory, isn't it?
(Port from EHR/MoreGameModes)


Pietrodjaowjao:
I think scene change is required to sync for all clients, but will test later

Pietrodjaowjao:
I noticed not everyone knows what "Disable vote ban system" option do, so it will now force disable this system

Tommy-XL:
I'm thinking of merging this for now, but we'll use it after v2.0.0
I will also add More Gamemodes in Readme

PR #961, opened 2024-05-31
Doppelganger Code Overhaul

D1GQ:
# Overview
Overhaul for Doppelganger to use similar mechanics as Dollmaster as requested from @0xDrMoe 
# Futures
```diff
+ | Code overhaul for Doppelganger.
+ | Swap Doppelganger role info to last victim role info.
+ | Swap Doppelganger info to last victim info on murder attempt for Sheriff.
+ | Swap Doppelganger info to last victim info for FortuneTeller.
+ | Display Doppelgangers fake name and real name to ghost.
+ | Set each Doppelganger victim's name to their real name to all ghost.
+ | Doppelganger victim's role will now be revealed to Doppelganger on kill.
```

### DIFF
--- a/Roles/Neutral/Doppelganger.cs
+++ b/Roles/Neutral/Doppelganger.cs
@@ -14,7 +14,6 @@ internal class Doppelganger : RoleBase
     public override CustomRoles ThisRoleBase => CustomRoles.Impostor;
     public override Custom_RoleType ThisRoleType => Custom_RoleType.NeutralKilling;
     //==================================================================\\
-
     private static OptionItem KillCooldown;
     private static OptionItem CanVent;
     private static OptionItem HasImpostorVision;
@@ -23,6 +22,7 @@ internal class Doppelganger : RoleBase
     public static readonly Dictionary<byte, string> DoppelVictim = [];
     public static readonly Dictionary<PlayerControl, byte> PlayerControllerToIDRam = []; // Edit ids!
     public static readonly Dictionary<byte, GameData.PlayerOutfit> DoppelPresentSkin = []; // Don't edit ids!
+    public static readonly Dictionary<byte, string> TrueNames = []; // Don't edit ids!
     public static byte CurrentIdToSwap = byte.MaxValue;
 
     public override void SetupCustomOption()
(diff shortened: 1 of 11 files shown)


Ultradragon005:
I think wait for merge of https://github.com/0xDrMoe/TownofHost-Enhanced/pull/963 , so doppelganger can implement late skinchange (allowing it to steal skins even in camoflague)

Ultradragon005:
> I think wait for merge of #963 , so doppelganger can implement late skinchange (allowing it to steal skins even in camoflague)

Oh wait I'd need to add nameplat and lvl rpc for that.. not bothered LOL

D1GQ:
> I think wait for merge of #963 , so doppelganger can implement late skinchange (allowing it to steal skins even in camoflague)

Later on what we could do is manually set the level and nameplate with a rpc then use the late skinchange for the actual cosmetic, it's not like the players can see the level and nameplate during game.

Roles/Neutral/Doppelganger.cs
```diff
@@ -73,6 +77,8 @@ public override void Add(byte playerId)
     // A quick check if a player has been killed by the doppelganger.
     public static bool CheckDoppelVictim(byte playerId) => DoppelVictim.ContainsKey(playerId);
 
+    public static PlayerControl GetDoppelControl(PlayerControl player) => DoppelgangerTarget != null ? DoppelgangerTarget : player;
```
Ultradragon005, line 80: btw did you test if storing playercontrol makes it possible to nullify it?

Like if the player leaves, does the reference (Doppelgangertarget in this case) to it's playercontrol also become null?

If not then a better way would to check if main.allplayercontrols.contains(doppelgangertarget), to know if it has left or not.

Gurge44: Storing classes with variables in C# keeps the same reference to that instance in all places. So yes, it will be null if they disconnect. (Same applies for lists, dictionaries, etc.)

D1GQ: @Ultradragon005 No I haven't tested it but knowing the problems with Dollmaster stored targets yes if the doppelganger does disconnect it will be null, but as you can see if it is null it just Returns the original controller.

Ultradragon005: lol gurge already confirmed my questions

PR #968, opened 2024-06-05
Modded Non-Host Mark fix + more

D1GQ:
# Overview
Fix a issue with Modded Non-Host clients marks not working and more!
# Futures
```diff
+ | Added RPC to receive players marks from the host.
+ | Added list of saved players marks.
```

### DIFF
--- a/Modules/RPC.cs
+++ b/Modules/RPC.cs
@@ -48,6 +48,8 @@ enum CustomRPC : byte // 194/255 USED
     RetributionistRevenge,
     SetFriendCode,
     SyncLobbyTimer,
+    RequestMarks,
+    ReceiveMarks,
 
     //Roles 
     SetBountyTarget,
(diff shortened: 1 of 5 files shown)


Ultradragon005:
uh.. there is like a WHOLE thing which this ignore
image

Ultradragon005:
also isn't playercontrol patch already for modded clients 
image

D1GQ:
> also isn't playercontrol patch already for modded clients image

Yes, I think you're missing the point that this is for modded clients that aren't the host.

Ultradragon005:
> Yes, I think you're missing the point that this is for modded clients that aren't the host.

playercontrol.localplayer is the moddedplayer

D1GQ:
> playercontrol.localplayer is the moddedplayer
> 
> It will activate for all players who have the mod installed and host

All modded players others than the host cannot get mark information because all of the roles are only ran through the host, so the GetMark override doesn't work. So to fix it we have to retrieve the information from the host.

Ultradragon005:
then it's better to do 

sendrpc:
writer.writnetobject{target};
writer.write($"{realname}{deathreason}{mark}{suffix}")


recievierpc:

target.cosmetics.nametext.text = reader.readstring();

D1GQ:
> sendrpc: writer.writnetobject{target}; writer.write($"{realname}{deathreason}{mark}{suffix}")
> 
> recievierpc:
> 
> target.cosmetics.nametext.text = reader.readstring();

I'll see what I can change tomorrow.

Tommy-XL:
Anyway fixed

PR #970, opened 2024-06-05
Delay & remove chats for green beans

NikoCat233:
If a player left and a message that sendto is that player, it will broadcast to everyone

This feature was introduced in spawn time out but got reverted
Now gonna return it

### DIFF
--- a/Patches/ChatCommandPatch.cs
+++ b/Patches/ChatCommandPatch.cs
@@ -2428,8 +2428,26 @@ class ChatUpdatePatch
         if (player == null) return;
         
         (string msg, byte sendTo, string title) = Main.MessagesToSend[0];
+
+        if (sendTo != byte.MaxValue)
+        {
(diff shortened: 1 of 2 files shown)


Patches/ChatCommandPatch.cs
```diff
@@ -2428,8 +2428,26 @@ public static void Postfix(ChatController __instance)
...
         
         (string msg, byte sendTo, string title) = Main.MessagesToSend[0];
+
+        if (sendTo != byte.MaxValue)
+        {
+            if (Utils.GetPlayerInfoById(sendTo) != null)
+            {
+                var targetinfo = Utils.GetPlayerInfoById(sendTo);
+
+                if (targetinfo.DefaultOutfit.ColorId == -1)
```
Tommy-XL, line 2438: -1 is fortegreen color?

Tommy-XL: Uhh, i see
image

PR #972, opened 2024-06-06
Players Know All Roles When Game End

Tommy-XL:
image

Note: does not always work, for example, when crewmates win, the player names are not displayed at all

image


NikoCat233:
You can just make the reason as Impostor win i guess

Tommy-XL:
> You can just make the reason sent to clients always as Impostor win i guess

Done, lol

PR #976, opened 2024-06-07
Make /s show the sender's name

NikoCat233:
1. Title
2. ClientData name is updated along with GameData name. Need to store it before gamedata got updated


Pietrodjaowjao:
i think it should be all in string name

NikoCat233:
Need to show their roles

PR #988, opened 2024-06-11
New PlayerControl function to get vent ID

D1GQ:
# Overview
Added a way to get the vent ID a player is currently in instead of just restricting it to Dollmaster!
## Futures:
```diff
= Change Dollmaster to use new method to get current vent ID.
+ Added new PlayerControl function to get vent ID that the player is currently in.
```

### DIFF
--- a/Modules/GuessManager.cs
+++ b/Modules/GuessManager.cs
@@ -379,6 +379,9 @@ public static class GuessManager
                 {
                     _ = new LateTask(() =>
                     {
+                        pc.GetRoleClass().OnGuessed(isUI, pc, target, role, ref guesserSuicide);
+                        target.GetRoleClass().OnGuessed(isUI, pc, target, role, ref guesserSuicide);
+                        
                         Main.PlayerStates[dp.PlayerId].deathReason = PlayerState.DeathReason.Gambled;
                         dp.SetRealKiller(pc);
                         RpcGuesserMurderPlayer(dp);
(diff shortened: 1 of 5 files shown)


Modules/GuessManager.cs
```diff
@@ -379,6 +379,9 @@ public static bool GuesserMsg(PlayerControl pc, string msg, bool isUI = false)
                 {
                     _ = new LateTask(() =>
                     {
+                        pc.GetRoleClass().OnGuessed(isUI, pc, target, role, ref guesserSuicide);
+                        target.GetRoleClass().OnGuessed(isUI, pc, target, role, ref guesserSuicide);
+
```
Tommy-XL, line 384: This is not needed because OnMurderPlayerAsTarget or OnMurderPlayerAsKiller already handles killing, you just need to use `inMeeting`

image

Tommy-XL: I mean `inMeeting` can be used to know that the player died because of a guesser or because he was juged, ect

D1GQ: I can remove the check on medic but isn't it better to be able to identify between a kill and a guess, for example you don't know if it's a counselor kill or is a guess.

Ultradragon005: > This is not needed because OnMurderPlayerAsTarget or OnMurderPlayerAsKiller already handles killing, you just need to use `inMeeting`

image
?

Ultradragon005: > I can remove the check on medic but isn't it better to be able to identify between a kill and a guess, for example you don't know if it's a counselor kill or is a guess.


Meeting kills are done via exile


But also this exists

image

Tommy-XL: image

D1GQ: Hmm I'll remove it and keep the vent ID function, I don't know what's up with medic because there's no logs lmao.

Tommy-XL: Damn, we have two functions that track murder, this is strange

PR #990, opened 2024-06-11
fix (💀)

Ultradragon005:
I forgot playercontrol.die is not an rpc. But it was working fine for me because I was testing as host lollll


Ultradragon005:
actually nvm I'm jsut hallucinating.

PR #993, opened 2024-06-12
Fixed interactions on protected players!

D1GQ:
# Overview
Fixes a long ongoing issue where protected players cannot be interacted with any Non-Killing roles!
## Futures:
```diff
+ Added new rolebase override to protect players from direct kills, but still allow interactions.
```

### DIFF
--- a/Patches/PlayerControlPatch.cs
+++ b/Patches/PlayerControlPatch.cs
@@ -283,6 +283,15 @@ class CheckMurderPatch
         if (killer.Is(Custom_Team.Impostor) && !Madmate.ImpCanKillMadmate.GetBool() && target.Is(CustomRoles.Madmate))
             return false;
 
+        Logger.Info($"Start", "OnCheckMurderAsTargetOnOthers");
+
+        // Check murder on others targets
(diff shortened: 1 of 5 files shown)


Ultradragon005:
this is very problematic.

Hence killers may use rpcmurderplayer in the direct kill (e.g swift or kamikaze). 

And RpcCheckAndmurder(check: true) is used standard to determine indirect kills.

Ultradragon005:
It would be better to rework the forcedcheckasmurder method to make that if return true, then at the end of checkmurder returns false (since it is only used for roles that have kill button ability). Or jsut make a seperate method which does that.

Ultradragon005:
> It would be better to rework the forcedcheckasmurder method to make that if return true, then at the end of checkmurder returns false (since it is only used for roles that have kill button ability). Or jsut make a seperate method which does that.

actually, jsut make admirer use forcedcheckmurder as killer.

But later (when new roles drop) I'll make a "OnInteracted(killer, target)" method.

Tommy-XL:
I think that if move the Admirer ability it will lead to another bugs, since the role literally conveverted to another team for example as a Cultist

Tommy-XL:
Therefore, I believe that there is no need to fix this

PR #995, opened 2024-06-12
Recruited Pelican Fix + More

D1GQ:
Fixed issue when jackal recruited Pelican that has eaten players. Also fixed text position issues when display role info is on for vanilla clients.

### DIFF
--- a/Modules/Utils.cs
+++ b/Modules/Utils.cs
@@ -1804,7 +1804,6 @@ public static class Utils
                 if (NameNotifyManager.GetNameNotify(seer, out var name))
                 {
                     SelfName = name;
-                    IsDisplayInfo = false;
                 }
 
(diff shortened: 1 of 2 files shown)


Tommy-XL:
Not the best way since the role uses two FixedUpdates at the same time

If Pelican changes role, then it is better to use Remove in this case
So I'll fix it

Modules/Utils.cs
```diff
@@ -1824,7 +1823,9 @@ static int GetInfoSize(string RoleInfo)
                 // Camouflage
                 if (!CamouflageIsForMeeting && Camouflage.IsCamouflage)
                     SelfName = $"<size=0%>{SelfName}</size>";
-
+                
+                if (!SelfName.Contains(seer.GetRealName()))
+                    IsDisplayInfo = false;
```
Tommy-XL, line 1828: I will transfer these changes myself too

PR #1008, opened 2024-06-15
Fix swift & berserker interaction(s)

Ultradragon005:


Ultradragon005:
bruh I have to think checkdoubletrigger trough

Ultradragon005:
now ready

Roles/Core/CustomRoleManager.cs
```diff
@@ -243,6 +252,15 @@ public static bool OnCheckMurder(ref PlayerControl killer, ref PlayerControl tar
                     DollMaster.CheckMurderAsPossessed(killer, target);
                     return false;
                 }
+
+        if (canceled)
+            return false;
+
+        if (cancelbutkill)
+        {
+            killer.RpcMurderPlayer(target);
```
Tommy-XL, line 261: But Swift will kill twice and teleport to the dead body

Ultradragon005: oh right, I forgot to change inside swift code 😅 fixed

PR #1026, opened 2024-06-19
Update some rpc calls

NikoCat233:
Update set outfit rpc
Update set role rpc (canOverrideRole is force set to true)
Disable eac until i rework it
Add basis for phantom

Playercontrol.hasimpkillbutton needs to be recoded for phantom
Pls add the new vanilla role base for the new roles


Pietrodjaowjao:
why

NikoCat233:
nvm just merge this first

PR #1028, opened 2024-06-20
Rename Masochist to Punching Bag

KingPanda360:
Masochist is a word that Innersloth has set to be an inappropriate word, so it would just appear as stars. Drakos had a good idea to rename it to "Punching Bag". This makes it so whenever people do /m or /r Punching Bag, the name will not appear as just stars.


Ultradragon005:
image
fix this tommy or water tomorrow

KingPanda360:
> image fix this tommy or water tomorrow

Fixed

PR #1037, opened 2024-06-24
Rename Schizophrenic to Paranoia and add role descriptions for AU vanilla roles

KingPanda360:
Schizophrenic has the word “Schizo,” which Innersloth has as an inappropriate word. I’m considering using “paranoia” as the new name because it's kind of similar to Schizophrenic.


Pietrodjaowjao:
innersloth probably being schizophrenic and thinkint someone would say this in a bad way

KingPanda360:
> innersloth probably being schizophrenic and thinkint someone would say this in a bad way

Then why would only the "Schizo" be the only thing blocked out if that was the case? There are other words that have "Schizo" in it besides Schizophernic

PR #1043, opened 2024-06-27
Disable Desync Roles From Getting ghost-roles

Ultradragon005:
Because innersloth thought it would be a funny idea to tease us with "CanOvverride" and never use it.


Gurge44:
"tought"

PR #1046, opened 2024-06-28
Fix rainbow being fucked up

NikoCat233:
Damn who code rainbow to update on every single player while not recoding the cool down


KingPanda360:
I know who it was but I really don't want to say who did it out loud.

PR #1052, opened 2024-06-30
Fix Phantom For Desync Roles

Tommy-XL:
As we know desync roles can see Phantom while invisible and in AU code only player have Phantom can make him invisible
To fix this we can use `RpcEnterVentDesync` and `RpcExitVentDesync` or `RpcBootFromVentDesync` only for desync roles

Very experimental at the moment


Tommy-XL:
Works extremely crappy because of the damn `lastSequenceId` for `SnapTo`

Why innersloth just not did teleport the player without fucking checking with `lastSequenceId`

PR #1059, opened 2024-07-05
/me stuff

Ultradragon005:
not perfect but works very well, and if any translation names are super long and they don't put <size=??%> </size> skill issue 🤷‍♂️

image


Pietrodjaowjao:
motels are different in my country..

PR #1068, opened 2024-07-09
 Delay the spawn for networked data for new clients

NikoCat233:
Networked data is huge, fat, rat. We must divide it so we wont disconnect anyone
image


NikoCat233:
SendAllStreamingNetObjects sends every net object with a single writer so we dont need to divide there.

PR #1070, opened 2024-07-09
/poll + QOL changes

Ultradragon005:
image
image
image


Ultradragon005:
Me personaly I FW ghadana

PHGaming121:
you should add a {time} to the end of the command, to change the duration of the poll
something like: /poll question A B C time (Example: /poll youre or you're? youre you're both 5 (the poll will end 5 mins after the same was posted in chat))

Ultradragon005:
at best I could make that a setting, but not a part of the command as that would overcomplicate things

Tommy-XL:
It's fully done?

Ultradragon005:
yeah

PR #1077, opened 2024-07-11
Dollmaster Fixes

D1GQ:
# Overview
More fixes for the beloved Dollmaster!
# Futures
```diff
+ Possibly fixed issue of players being swapped during meeting.
+ Fixed issue possessing player during comms camouflage sabotage.
+ Fixed desync issues with swapping players cosmetic.
+ Fixed Dollmaster color showing on map for possessed player.
+ Fixed possessed player being able to see self/dollmaster outfit inside a fungal mushroom cloud.
+ Fixed vanilla info diplay in Camouflage Comms.
```

### DIFF
--- a/Modules/Utils.cs
+++ b/Modules/Utils.cs
@@ -1824,6 +1824,13 @@ public static class Utils
                         SelfName = GetString("DevouredName");
                 }
 
+                // Dollmaster, Prevent seeing self in mushroom cloud
+                if (CustomRoles.DollMaster.HasEnabled() && seerRole != CustomRoles.DollMaster)
+                {
+                    if (DollMaster.IsDoll(seer.PlayerId))
+                        SelfName = "<size=10000%><color=#000000>■</size></color>";
+                }
+
                 // Camouflage
                 if (!CamouflageIsForMeeting && Camouflage.IsCamouflage)
                     SelfName = $"<size=0%>{SelfName}</size>";
--- a/Modules/Utils.cs
+++ b/Modules/Utils.cs
@@ -1828,12 +1828,15 @@ public static class Utils
                 if (CustomRoles.DollMaster.HasEnabled() && seerRole != CustomRoles.DollMaster)
                 {
                     if (DollMaster.IsDoll(seer.PlayerId))
-                        SelfName = "<size=10000%><color=#000000>■</size></color>";
+                        SelfName = "<size=10000%><color=#000000>■</color></size>";
                 }
 
                 // Camouflage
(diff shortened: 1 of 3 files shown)


Ultradragon005:
I swear how many  bug fix prs is this role gonna have DAMN 💀

D1GQ:
True lmao, To be fair Dollmaster is a bit more complex than most roles.

Dxgamer7405:
I thought about something, what if the tracker tracks the dollmaster? he will probably be able to see who is controlling him that way

D1GQ:
> I thought about something, what if the tracker tracks the dollmaster? he will probably be able to see who is controlling him that way

This shouldn't be the case due to it now keeping the PlayerNetworkedData the same. Now it will be tracking the wrong player but there's no way to fix that at the time!

Modules/Utils.cs
```diff
@@ -1824,6 +1824,13 @@ static int GetInfoSize(string RoleInfo)
                         SelfName = GetString("DevouredName");
                 }
 
+                // Dollmaster, Prevent seeing self in mushroom cloud
+                if (CustomRoles.DollMaster.HasEnabled() && seerRole != CustomRoles.DollMaster)
+                {
+                    if (DollMaster.IsDoll(seer.PlayerId))
+                        SelfName = "<size=10000%><color=#000000>■</size></color>";
```
Tommy-XL, line 1831: html code is written incorrectly
Should be:
`"<size=10000%><color=#000000>■</color></size>"`

Patches/PlayerControlPatch.cs
```diff
@@ -1325,6 +1325,13 @@ public static Task DoPostfix(PlayerControl __instance)
                         RealName = GetString("DevouredName");
                 }
 
+                // Dollmaster, Prevent seeing self in mushroom cloud
+                if (CustomRoles.DollMaster.HasEnabled() && seerRole != CustomRoles.DollMaster)
+                {
+                    if (DollMaster.IsDoll(seer.PlayerId))
+                        RealName = "<size=10000%><color=#000000>■</size></color>";
```
Tommy-XL, line 1332: Same

Modules/Utils.cs
D1GQ: I'll fix it, but how's that incorrect? I've always written html code like that in among us and never had an issue.

Ultradragon005: In the provided snippet, it works as expected. it's just not good practice to begin closing the outer-most tags BEFORE closing the inner-most tags
image

PR #1081, opened 2024-07-12
Add blacklisting of some roles for mini + fix bug

LezaiYa1:


NikoCat233:
@Tommy-XL should we just patch all check murder attempts on mini directly

Pietrodjaowjao:
> @Tommy-XL should we just patch all check murder attempts on mini directly

Why

PR #1085, opened 2024-07-14
Rework Councillor Conditons

NikoCat233:
I think someone just copy and paste from judge when coding this role,
with litter brain in head.

Add setting to reveal Councillor Trail ( from judge trail)
Removed guess obvious role limits for Councillor since its not a guesser
Add medic shield and councillor no longer suidice for judging those with a shield
Add settings to prevent judging task done snitch
Judge and Councillor now can not trail those protected by medic
Converted councillor can also judge freely
Silently added my alt to devmanager


LezaiYa1:
CV code

Gurge44:
Guess who copy-pasted the code....
(yes, you're right)

Ultradragon005:
> Guess who copy-pasted the code.... (yes, you're right)

I think that might've been Gerald actually

PR #1088, opened 2024-07-15
Fix butcher realted

NikoCat233:
Like youtuber, revenge butcher kills its not shown in kill logs some times

### DIFF
--- a/Roles/Impostor/Butcher.cs
+++ b/Roles/Impostor/Butcher.cs
@@ -1,4 +1,5 @@
-﻿using TOHE.Roles.Core;
+﻿using TOHE.Modules;
+using TOHE.Roles.Core;
 using UnityEngine;
 using static TOHE.Options;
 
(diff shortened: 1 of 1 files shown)


Roles/Impostor/Butcher.cs
```diff
@@ -55,11 +58,12 @@ public override void OnMurderPlayerAsKiller(PlayerControl killer, PlayerControl
             var pcList = Main.AllAlivePlayerControls.Where(x => x.PlayerId != target.PlayerId); //No need to do extra check cause nobody is winning
             pcList.Do(x =>
             {
+                Main.PlayerStates[x.PlayerId].IsDead = true;
```
Tommy-XL, line 61: Better use `.SetDead();` becouse modded clients not synced

NikoCat233: I just dont want to send rpc here lol

Tommy-XL: Hmm

PR #1098, opened 2024-07-18
Add new roles in alphabetical order in CustomRole Enum file

GameTechGuides:
Add new roles in alphabetical order in the `CustomRole` Enum file in `main.cs`.

* Add new roles such as `Admirer`, `Alchemist`, `Addict`, etc. (all of them), in alphabetical order.
* Ensure roles are listed alphabetically within their respective categories.
* Maintain the grouping of roles by categories such as Crewmate, Impostor, Neutral, etc.

### DIFF
--- a/main.cs
+++ b/main.cs
@@ -534,29 +534,29 @@ public enum CustomRoles
 
     // Crewmate(Vanilla)
     Crewmate = 0,
-    Scientist,
     Engineer,
     GuardianAngel,
-    Tracker,
     Noisemaker,
+    Scientist,
+    Tracker,
 
     // Impostor(Vanilla)
     Impostor,
-    Shapeshifter,
     Phantom,
+    Shapeshifter,
 
     // Crewmate Vanilla Remakes
     CrewmateTOHE,
-    ScientistTOHE,
     EngineerTOHE,
     GuardianAngelTOHE,
-    TrackerTOHE,
     NoisemakerTOHE,
+    ScientistTOHE,
+    TrackerTOHE,
 
     // Impostor Vanilla Remakes
     ImpostorTOHE,
-    ShapeshifterTOHE,
     PhantomTOHE,
+    ShapeshifterTOHE,
 
     // Impostor Ghost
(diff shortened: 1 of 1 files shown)


GameTechGuides:
Guys, sometimes I miss things. If I forget to add something in alphabetical order, please let me know.

Ultradragon005:
I mean it rlly dosen't matter, ryuk just forgot to remove his comment

KingPanda360:
Specter is renamed TOHE Phantom before v2024.6.18 came out and added the new vanilla role called "Phantom". It was renamed to Spectre because there would be two roles of the same name. This should not be removed from the list.

Ultradragon005:
> Specter is renamed TOHE Phantom before v2024.6.18 came out and added the new vanilla role called "Phantom". It was renamed to Spectre because there would be two roles of the same name. This should not be removed from the list.

Oh wait I just realized... wtf... 💀 how old must this branch be.

Also it takes 2 seconds to do this with chatgpt if  we would like to

main.cs
```diff
@@ -746,14 +745,13 @@ public enum CustomRoles
     Opportunist,
     Pelican,
     Pestilence,
-    Specter,
```
Tommy-XL, line 749: Specter missed

GameTechGuides: @Tommy-XL, thanks for pointing out that Spectre was missed. I'll correct that immediately.

Ultradragon005:
[PDFen (1).pdf](https://github.com/user-attachments/files/16287782/PDFen.1.pdf)

The positions don't rlly matter, but if doing the whole thing might aswell go all out, so by asking gpt I got the exact positions (which take into account their second-msot, third-most fourth-msot letter etc..). (left side the current position, right side the most correct position)

Pietrodjaowjao:
when this pr is merged i will create one randomizing the order completely

Ultradragon005:
> when this pr is merged i will create one randomizing the order completely

yoooooo

PR #1103, opened 2024-07-20
Add Scanner for overseer

Ultradragon005:
visualtasks or not dosen't matter because in the rpc it dosen't check for that, only in innersloth's own method it does


Gurge44:
Yeah Herbalist in PL also used this

Tommy-XL:
I hope this doesn't cause disconnects when constantly sending rpc lol

Ultradragon005:
> I hope this doesn't cause disconnects when constantly sending rpc, lol

uh... have you like..... seen glitch or chronomancer 💀 they're spamming changenamerpc almost every second

Ultradragon005:
and seems to be perfectly fine 😼

PR #1110, opened 2024-07-22
Just Enable EAC

NikoCat233:
Enable Playercontrol part EAC
Fix checkDenyName not working
Change some check conditions

**This pr needs further testing.**
Since host is using custom rpc sender or some StartRpcImmediately, which may set the target to host itself
And server will return the rpcs to host, resulting in false reports for host and other clients.
SetName and SetRole is sent by host and also received by host itself. I have to disable them.
Not sure whether there are other rpc than can be sent and received by host itself. So we need testing

Note that all the checks related to EAC can be spoofed by hackers with other's identity easily
~~**The main goal of EAC is no longer to stop cheat since it basically can't. Only dumb hackers can be caught. So it is there just to prevent further destruction to the room.**~~
If you really dont want to be hacked, enable white list on official servers, do dm for code or just go to Niko's anticheat region


LezaiYa1:
我也试过这样改，但还是防不住外挂在大厅炸管=(

NikoCat233:
Not working

Dxgamer7405:
omg

PR #1114, opened 2024-07-24
Enhanced EAC detection of AUM

LezaiYa1:


NikoCat233 (changes requested):
Lobby vent is already checked in player physics eac

Lobby timer is sent by lobby behaviour and we cant know who exactly sends the rpc like ship status. This can only be patched server side since innersloth is also sending it. There is always a reason where i know a cheat but i dont tell or fix it in eac. Impostor probably gonna include anticheat regarding lobby behaviour after 6.18 release.

These are both not sent by player control.
You didnt fully understand the protocol and make it wrong lol.

Maybe i have to speed up my process for the among us protocol video

NikoCat233:
And aum cant do above cheats

PR #1118, opened 2024-07-26
Simple PhantomBaseRole Support

D1GQ:
# Overview
Added a bit of support for Phantom Base Roles for testing, it will probably be a while before any actual Phantom Roles come out.
# Futures
```diff
+ Added new RPC function for Appear & Vanish called Phantomize.
+ Added RoleBase for Phantom ability.
+ Added ability to disable animation for Vanish.
===  Not originally possible for vanilla, and not the best solution, so mostly temporarily until a better way is found for this and desync imp roles!
+ Added utility to get furthest vent from all alive players.
+ Added new check on ExtendedPlayerControl called IsInShapeshift
+ Added new check on ExtendedPlayerControl called IsInVanish
```


Patches/PlayerControlPatch.cs
```diff
@@ -725,6 +725,76 @@ class CheckVanishPatch
...
+        {
+            return false;
+        }
+
+        // No called code if is invalid vanish
+        if (!CheckInvalidVanish(__instance))
+        {
+            return false;
+        }
+
```
Tommy-XL, line 739: Returning false doesn't make any sense because vanilla players will constantly play the disappearing animation

Tommy-XL: Therefore, the phantom should have a `void` function and not a `bool`

Pietrodjaowjao:
This is not really needed anymore

Patches/PlayerControlPatch.cs
D1GQ: Doesn't Check get sent by the vanilla clients to confirm vanish and appear is valid, how come making it return false wouldn't do anything?

Tommy-XL: Try to cancel CheckVanish on vanilla client and see what happens

D1GQ:
About to go ahead and close this PR since it's already been done!

PR #1119, opened 2024-07-27
Rename Anti-Adminer to Cyberpunk

KingPanda360:
The Anti-Adminer needs a better name, 100%. Cyberpunk is a fantastic name because it's similar to what the Anti-Adminer does: hack into devices to find out when people are using them. In the past, KARPED1EM even said that this role needs a better name.


NotPyro404:
I don't think the name Cyberpunk is good for this role. Maybe something like Interference, not sure.

KingPanda360:
> I don't think the name Cyberpunk is good for this role. Maybe something like Interference, not sure.

I think interference isn't a good word to use as a role name because it seems too simple and broad. Do you have any other name ideas?

NotPyro404:
> I think interference isn't a good word to use as a role name because it seems too simple and broad. Do you have any other name ideas?

Hmm, not really sadly. Maybe Throttler because connection throttled (but that kind of doesn't make sense, I don't know what would be like evil telecommunication)

KingPanda360:
> > I think interference isn't a good word to use as a role name because it seems too simple and broad. Do you have any other name ideas?
> 
> Hmm, not really sadly. Maybe Throttler because connection throttled (but that kind of doesn't make sense, I don't know what would be like evil telecommunication)

Cyberpunk may sound like a bad role name to you, but it fits the description of Anti-Adminer, plus some other people agree with my suggestion. So that's why I made this.

NotPyro404:
> Cyberpunk may sound like a bad role name to you, but it fits the description of Anti-Adminer, plus some other people agree with my suggestion. So that's why I made this.

I guess that's fair.

Tommy-XL:
> I don't think the name Cyberpunk is good for this role. Maybe something like Interference, not sure.

I agree the title is fundamentally inappropriate for this role, and why change it at all?

NotPyro404:
> I agree the title is fundamentally inappropriate for this role, and why change it at all?

Anti-Adminer isn't a great name either, I would prefer a 3rd candidate instead.

Ultradragon005:
GUYS, I have. The perfect name, call it "Blotnik"

KingPanda360:
> Blotnik

Why this?

Ultradragon005:
> Why this?
BLOTNIK!!!!
image

KingPanda360:
> > I don't think the name Cyberpunk is good for this role. Maybe something like Interference, not sure.
> 
> 
> 
> I agree the title is fundamentally inappropriate for this role, and why change it at all?

I wanted to change it because Anti-Adminer didn't seem like a good name for a role. I wanted to make a role name simple and easier to type, say, and spell.

Ultradragon005:
image
good boy

Tommy-XL:
Sorry, lol

image

Ultradragon005:
Oh My goodnes, see. This is why you should've named the PR "Rename Anti-Adminer to Blotnik" and then everything would be fine

PR #1132, opened 2024-07-29
Change API connect system

Tommy-XL:
Removes disconnect when the API completely down in Canary and Release builds


Ultradragon005:
Finally

PR #1151, opened 2024-08-11
Optimize code

Tommy-XL:
Some code is very complex, so it's better to simplify it

Note: Noticed some wrong changed some code in, lol


Tommy-XL:
I think i open new

PR #1157, opened 2024-08-13
Fix Egoist win text

KingPanda360:


Ultradragon005:
and what if multiple egost wins 🤨

Ultradragon005:
> Is that possible? I thought there could only be one egoist in a game.
it is. also in the winnercheck code it cheks for multiple egoists of the same team
image

Ultradragon005:
I mean you can just do Egoist(s) Win!

KingPanda360:
I was thinking that, but I was wondering if people would get confused by it.

Ultradragon005:
there could be 2 texts, but idk who wants top code that

KingPanda360:
> there could be 2 texts, but idk who wants to code that

If someone wants to do that, they could make a separate PR, and this one can be closed without merging.

PR #1161, opened 2024-08-14
Fix Textboxpatch and Fix SetupHelpIcon

Ultradragon005:
so they jsut changed the passive buttron to a gameoptionbuttons which inherits it lol (but it does not have all features, such as active/inactive sprite is missing it instead stores it as colors)


Ultradragon005:
also alpha 4 is very broken LOL

PR #1162, opened 2024-08-14
New Imposter Ghost Role: Possessor

D1GQ:
# Possessor
As the Possessor, you are able to possess players when others aren't in the Alert Range. Lead the possessed player as far as possible from other players that are in the Focus Range. Once the possession duration is up, the possessed player will be killed if others aren't in the Focus Range. If you run into another player in the Alert Range while possessing, the Possessor will immediately unpossess.


Roles/(Ghosts)/Impostor/Possessor.cs
```diff
@@ -0,0 +1,167 @@
...
+    //==================================================================\\
+    public static bool controllingPlayer = false;
+    public static byte controllingTargetId = byte.MaxValue;
+    public static float controllingLastSpeed = float.MinValue;
+    public static float possessTime = float.MinValue;
+
+    public static OptionItem PossessCooldown;
+    public static OptionItem PossessDuration;
+    public static OptionItem AlertRange;
+    public static OptionItem FocusRange;
```
Tommy-XL, line 25: Set `private` If this is not used outside of the role class

Roles/(Ghosts)/Impostor/Possessor.cs
```diff
@@ -0,0 +1,167 @@
...
+                    if (!target.petting && CheckRange(_Player.GetCustomPosition(), target.GetCustomPosition()) > 1f)
+                    {
+                        target.MyPhysics.RpcPet(_Player.GetCustomPosition(), new Vector2(500f, 500f));
+                    }
+                    else if (!target.MyPhysics.Animations.IsPlayingRunAnimation())
+                    {
+                        target.MyPhysics.RpcCancelPet();
+                        target.RpcTeleport(target.GetCustomPosition());
+                    }
+                }
```
Tommy-XL, line 92: I don't quite understand what this code is

Why does the player need to go to coordinates x 500 y 500

Ultradragon005: He manipulates Petrpc to make the player walk smoothly towards the direction, instead of spamming teleport on the player https://discord.com/channels/1094344790910455908/1270477748443283517/1270477748443283517

PR #1183, opened 2024-08-23
Fix AntiBlackOut Rpc releated bugs

NikoCat233:
AntiBlackOut can be happening while we are still in lobby.
The old code is trash, it ends the game even when you are in lobby.
Leading to this issue. https://github.com/Impostor/Impostor/issues/646

Explained Changes:
1. Stop the game from starting if we are still in lobby and kick the f*cked player
2. Depending on the settings, we directly end the game or try to kick the f*cked player and continue the game
3. No matter the case, the f*cked player should be safely kicked.
4. Now the game ends immediately if the blackout rpc is received after Intro being displayed

**Important : I didnt test the codes myself. Someone Please Review and test it before merging**
It can be easily tested by manually sending AntiBlackOut Rpc as host & clients in different conditions mentioned above


NikoCat233:
Also this should be tested on both modded regions and official servers. They behave differently regarding end game calls

NikoCat233:
Tested on Modded Region, should be working

PR #1182, opened 2024-08-21
Role Basis Changer (port from TOH-Y)

Ultradragon005:
It's F#cking awesome

- Change role midgame
- Revive Players
- STABLE INVISIBILITY FOR EVERYONE, I think
- Completely fix blackscreen

TaskList:

- [x] Rigorously test it for any bugs/issues
- [x] Make A  proper RPCsetrole extension  method for making desync or switching player to imp team.
- [x] Completely fix blackscreen, temporarily remove resetcam (cuz innerdick might remove canovverride 😭). 
- [x] Desync players can become ghost roles again 😼

- [x] Make it so desync crewmates get assigned the impostor role after intro screen to make it much less confusing.
- Later, Make phantomrpc usable for everyone, by use rpcsetrole instead of vent desync. (I think this should theoretically be possible, but I would need to test this later to make sure)

PR #1186, opened 2024-08-24
Disable Use Vent For Vanilla (Port from MoreGamemodes)

Tommy-XL:
It seems to work, but I still don’t understand the structure of the code, and it seems there is an option to block certain ventilations

I also don't yet know how to return use vent again

image


Ultradragon005:
fr

Ultradragon005:
as ridicolous as it sounds, maybe you can reassign their roletype, by #1182 (make a new branch were u merge those 2 to test).
And see if that maybe resets it.
Unless rabek knows another way

Tommy-XL:
I understand how the code works
In fact, if vens is active for player again, need to send `pc.RpcSetVentInteraction()`

Ultradragon005:
oh okay, I noticed that but didn't think it would work, since the original uses Detoriate and not deserialize

Tommy-XL:
Damn, for some roles need to send an Rpc with a check for activate/blocking vents

NikoCat233:
It is working but part of the code is not quite good.
In vanilla ship status is normally updated with messagewriter instead of serializing data.
This code seems to be changing the data for overall ship status, which is unnecessary.
I will try code a better one later

NikoCat233 (changes requested):
I will try do some changes
Gonna open a pr regarding this late today

Rabek009:
NikoCat233 you don't have to code it. There is already better code in More Gamemodes.

Rabek009:
My code doesn't patch shipstatus. It only patch ventilation system.

NikoCat233:
> My code doesn't patch shipstatus. It only patch ventilation system.

https://github.com/Rabek009/MoreGamemodes/blob/5e2ad5467eb9931a0f4235708a9bc82b7d657b82/Patches/ShipStatusPatch.cs#L123

Obviously it is patching overall ship status serializing

And I suggest you check this change
https://github.com/0xDrMoe/TownofHost-Enhanced/pull/1195/files#diff-f28d7cf36b36f9eedd77ae1497343146c6ec19aa116ee0b8b2bdefce37da437aR636

Patches/VentSystemPatch.cs
```diff
@@ -0,0 +1,130 @@
...
+
+namespace TOHE.Patches;
+
+// Patches here is also activated from ShipStatus.Serialize and IntroCutScene 
+// through Utils.SetAllVentInteractions
+
+[HarmonyPatch(typeof(VentilationSystem), nameof(VentilationSystem.Deteriorate))]
+static class VentSystemDeterioratePatch
+{
+    public static Dictionary<byte, int> LastClosestVent;
```
NikoCat233, line 12: UnUsed variable. Will recode a better version of serialize.

NikoCat233 (changes requested):

Patches/HudPatch.cs
```diff
@@ -119,7 +119,7 @@ public static void Postfix(HudManager __instance)
                     __instance.KillButton.ToggleVisible(false);
                 }
 
-                bool CanUseVent = player.CanUseImpostorVentButton();
+                bool CanUseVent = player.CanUseVent();
                 __instance.ImpostorVentButton.ToggleVisible(CanUseVent);
```
Tommy-XL, line 123: Found bug, Impostor vent button will be visible for engineer based (for modded)

NikoCat233: it should not be changed here i think

Tommy-XL:
Now need to add `RpcSetVentInteraction` for those roles that temporarily block their vents, for example Arsonist, etc.
I'll do this later

PR #1187, opened 2024-08-24
Fixed Role Basis Changer (By: Drakos) (Port from TOH-Y)

Tommy-XL:


Tommy-XL:
Damn it was hard but the role assignment seems to be working fine

I don’t know about the change in roles, but everything seems fine there too

There is also a problem when Amnesiac or another role that copies the role of the dead checks the current role and not the one that was before his death

That is, if a dead player has a ghost role and the amnesiac tries to copy the role, he will not succeed

Tommy-XL:
Although, I was mistaken, he will be able to copy the role, but will assign the default role to a specific team

Ultradragon005:
what's the custom role gonna be sued for anyways in the rolemap

Tommy-XL:
> what's the custom role gonna be used for anyways in the rolemap

I don’t remember, but it will be useful to use it to remember the role that was originally
That's why it's a "role map"

Modules/AntiBlackout.cs
```diff
@@ -236,6 +258,45 @@ public static void AfterMeetingTasks()
...
+            var target = Utils.GetPlayerById(targetId);
+            
+            if (seer == null || target == null) continue;
+            if (seer.IsModClient()) continue;
+
+            var realRoleType = roletype;
+            if (target.Data.IsDead)
+            {
+                realRoleType = target.CanUseSabotage() ? RoleTypes.ImpostorGhost : RoleTypes.CrewmateGhost;
+            }
```
Tommy-XL, line 280: Note to self: This part of the code can be improved and immediately assign the Guardian Angel role there if necessary

Patches/MeetingHudPatch.cs
```diff
@@ -368,7 +368,7 @@ public static bool Prefix(MeetingHud __instance)
 
             ExileControllerWrapUpPatch.AntiBlackout_LastExiled = exiledPlayer;
             Main.LastVotedPlayerInfo = exiledPlayer;
-
+            if (exiledPlayer != null) AntiBlackout.ExilePlayerId = exiledPlayer.PlayerId;
```
Tommy-XL, line 371: Also when Dictator exiles a player, `ExilePlayerId` remains - 1

Tommy-XL:
Role assignment is broken, needs to be improved

Tommy-XL:
Role assignment is broken again, bruh

PR #1188, opened 2024-08-25
Better AntiBlackout

Ultradragon005:
set a dummy person to imp,
And revive everyone else (except for host, cuz host needs to know when to end the game).

And bring them back to normal afterwards.


Ultradragon005:
tbh, this fix is litteraly so good, that even if by miracle someone blackscreens, they unblackscreen the following meeting.

PR #1195, opened 2024-08-27
Improvements for DisableUseVentForVanilla

NikoCat233:
the old methods delays other dirty system types from updating when ventilation is dirty
Though it has little influence on real game play, i still decide to patch shipstatus like what we have done to Networked PlayerInfo

1. Move Serialize patch of shipstatus to shipstatus.cs and improve it so it wont delay other system types from updating
2. CanUseVent used to use player.data.role.role == engineer, obviously it sucks. Changed to GetVNRole
3. In IntroCutscene the old code stop adding LastVentId when it meets a "true". fixed that so last vent id is added to all needed players
4. Before IntroCutscene got destoryed (game really starts), the old code is serializing Vents (before role get assigned). This may make the first 2 vents in Skeld unusable. Fixed by delaying updating the custom vents until intro is being played

### DIFF
--- a/Modules/ExtendedPlayerControl.cs
+++ b/Modules/ExtendedPlayerControl.cs
@@ -680,7 +680,7 @@ static class ExtendedPlayerControl
             _ => false
         };
     }
-    public static bool CanUseVent(this PlayerControl player) => player.CanUseImpostorVentButton() || player.Data.Role.Role == RoleTypes.Engineer;
+    public static bool CanUseVent(this PlayerControl player) => player.CanUseImpostorVentButton() || player.GetCustomRole().GetVNRole() == CustomRoles.Engineer;
     public static bool CanUseImpostorVentButton(this PlayerControl pc)
     {
         if (!pc.IsAlive()) return false;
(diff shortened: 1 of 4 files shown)


Modules/ExtendedPlayerControl.cs
```diff
@@ -680,7 +680,7 @@ public static bool HasKillButton(this PlayerControl pc)
             _ => false
         };
     }
-    public static bool CanUseVent(this PlayerControl player) => player.CanUseImpostorVentButton() || player.Data.Role.Role == RoleTypes.Engineer;
+    public static bool CanUseVent(this PlayerControl player) => player.CanUseImpostorVentButton() || player.GetCustomRole().GetVNRole() == CustomRoles.Engineer;
```
NikoCat233, line 683: @Tommy-XL 
Can you check is this change is correct?

NikoCat233:
Tested in local clients and everything seems fine

Modules/ExtendedPlayerControl.cs
Tommy-XL: Yes, but it mostly doesn't make sense since we never use desync engineer

NikoCat233: Welp this is sussy
The host checks this and it works for vanilla
But im not sure whether previous code is ok for modded clients

So just change it since it doesnt cause bugs

PR #1196, opened 2024-08-30
Cache RpcCloseVent

NikoCat233:
Previously RpcCloseVent is sent by host at every fixed update to every block vent players
I add some codes to send the close vent only when the closest vent to a player changed.

Sorry I merged latest dev 2.1.0 into my branch. Just check the last commit

Known issues:
if players in a game is too little (like only 2-3 players)
then the amount of vents that can be closed to a player is the player count in that game.
So if we get too little players, some may still be able to vent before host could update the close vent calls


Tommy-XL:
Can you sync with last commit?

PR #1204, opened 2024-09-07
Altruist can't be Oblivious

KingPanda360:


Tommy-XL:
I move this code in my changes in Role Basics

PR #1212, opened 2024-09-19
Better Among Us RPC support

D1GQ:
# Added support for the Better Among Us RPC
- When a player is a BAU user their name will be green and will not be kicked for sending a untrusted RPC!
- This is added because i am currently working on support for BAU users being able to be in TOHE lobbies

### DIFF
--- a/Modules/RPC.cs
+++ b/Modules/RPC.cs
@@ -74,7 +74,7 @@ enum CustomRPC : byte // 193/255 USED
     SetLoversPlayers,
     SetExecutionerTarget,
     RemoveExecutionerTarget,
-    BetterCheck, // BetterAmongUs (BAU) RPC, This is sent to allow other BAU users know who's using BAU!
+    BetterCheck = 150, // BetterAmongUs (BAU) RPC, This is sent to allow other BAU users know who's using BAU!
     SendFireworkerState,
     SetCurrentDousingTarget,
     SetEvilTrackerTarget,
--- a/Modules/RPC.cs
+++ b/Modules/RPC.cs
@@ -22,6 +22,7 @@ enum CustomRPC : byte // 193/255 USED
     VersionCheck = 80,
     RequestRetryVersionCheck = 81,
     SyncCustomSettings = 100, // AUM use 101 rpc
+    BetterCheck = 150, // BetterAmongUs (BAU) RPC, This is sent to allow other BAU users know who's using BAU!
     SetDeathReason = 102,
     EndGame,
     PlaySound,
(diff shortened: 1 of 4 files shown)


Modules/RPC.cs
```diff
@@ -74,6 +74,7 @@ enum CustomRPC : byte // 193/255 USED
     SetLoversPlayers,
     SetExecutionerTarget,
     RemoveExecutionerTarget,
+    BetterCheck, // BetterAmongUs (BAU) RPC, This is sent to allow other BAU users know who's using BAU!
```
NikoCat233, line 77: Set this to a fixed value in case we add other custom rpcs in the future and change the callid

NikoCat233 (changes requested):

NikoCat233 (approved):

Pietrodjaowjao:
i recommend also changing the order to the top

NikoCat233:
no. this will make the final rpc number exceed 256

Pietrodjaowjao:
then bottom

NikoCat233:
You cant. It must be put before 150 so it is assigned 150

Tommy-XL:
I think I know what Pietro is talking about, if some Rpc is added above, then all the others below will shift the number by +1 and when they reach 150 they will not shift the number further but will remain at 151,152 and so on, so it is better to put fixed Rpc at the very top, for example after 103

Modules/RPC.cs
```diff
@@ -22,6 +22,7 @@ enum CustomRPC : byte // 193/255 USED
     VersionCheck = 80,
     RequestRetryVersionCheck = 81,
     SyncCustomSettings = 100, // AUM use 101 rpc
+    BetterCheck = 150, // BetterAmongUs (BAU) RPC, This is sent to allow other BAU users know who's using BAU!
```
Tommy-XL, line 25: I mean set `BetterCheck` after `SetDeathReason` and set 103 number, lol

NikoCat233:
he just want to set it to 150, same with the bau itself

D1GQ:
The BAU RPC is 150 and it's always been that since the start, i don't want to just change it because of compatibility reasons for other mods, and if I changed it, it would cause issues with the anti-cheat when people are using different versions of BAU

Modules/RPC.cs
D1GQ: I moved it up to the top so it doesn't affect any other RPC, but other than that there's not much I can do, it would cause issues if I change the RPC number from 150

PR #1221, opened 2024-09-22
Death Ejection Message

MargaretTheFool:
ejection messages arent showing at all even when I have it set to role but I'm gonna assume that's not an issue relating to me because I went through the logs and the code DOES run


Tommy-XL:
The message is not displayed because after everyone has voted you need to wait 4-5 seconds for the player's name to change and return to normal

PR #1256, opened 2024-10-06
Miner can't get Circumvent

KingPanda360:


Pietrodjaowjao:
no u just added the options back but they will not apply to anything

KingPanda360:
> no u just added the options back but they will not apply to anything

Oh I see. I might as well just leave it in Tommy's hands because I would need to add back all the stuff Drakos deleted for sleuth if I wanted to fix that. But tommy will have another way to do it.

PR #1260, opened 2024-10-09
Unhide Team Guessing

MargaretTheFool:


Tommy-XL:
Are these settings used in the code?

MargaretTheFool:
i know apoc guess apoc is because I coded that myself, I think the other two are used because I probably looked at those to code apoc guess apoc

Tommy-XL:
image

These settings are only used for modded players, but in the meantime the vanilla code does not check these settings, so they are simply partially unused

MargaretTheFool:
ok NOW they are

PR #1267, opened 2024-10-20
TOH send at 500

NikoCat233:
TOH send at 500
so we should change to 500


0xDrMoe:
why? we can just keep length cuz of formatting

NikoCat233:
for some reason current sending 2 inner playerinfo can still lead to packets above 1200 in size
so its better to follow toh that sends every 500

Tommy-XL:
Then it's better to add it later

PR #1273, opened 2024-10-27
Improve Guessing TNA

MargaretTheFool:
being unable to guess is now tied to the role not the person with the role


Tommy-XL:
Damn, i forgot change to main

PR #1286, opened 2024-11-09
Add Revenant (Neutral)

Limeau:
Idea: https://discord.com/channels/1094344790910455908/1304666536841056276

As the Revenant, your goal is to be killed. If you are killed, you will take your killer's role and kill your killer instead. You cannot win before being killed.


laikrai (commented):
I think you need to edit OptionHolder.cs line 636

Roles/Neutral/Revenant.cs
```diff
@@ -0,0 +1,34 @@
...
+    public override bool OnCheckMurderAsTarget(PlayerControl killer, PlayerControl target)
+    {
+        CustomRoles role = killer.GetCustomRole();
+
+        killer.RpcMurderPlayer(killer);
+        killer.SetRealKiller(target);
+
+        target.RpcChangeRoleBasis(role);
+        target.RpcSetCustomRole(role);
+
```
NikoCat233, line 29: target.getroleclass()?.onadd()

NikoCat233 (changes requested):

PR #1287, opened 2024-11-15
eac upgrade

redphantom1000:
added detections for KN because frick you dumb hackers. A bit unsure if this conflicts with the judge/councilor or not but when I tested this i had no problems


NikoCat233:
Cant confirm whether this is working. I dont have a sample of Kill Network

NikoCat233 (changes requested):
Cant confirm

NikoCat233 (approved):

PR #1295, opened 2024-11-25
Dev 2.2.0 translation

NikoCat233:


NikoCat233:
Seems like i messed up

PR #1296, opened 2024-11-25
Scavenger custom death reason (Option)

ApeMV:
Has been playtested, not much more to say.


ApeMV:
Not sure why it automatically wants to try adding the old PR's again. Let me know if this is a problem.

PR #1307, opened 2024-12-02
Fix being able to set more than 15 players in vanilla regions

Pietrodjaowjao:
This will cancel the command if the input is bigger than 15 and its in a vanilla region


Pietrodjaowjao:
i have no idea why it added so many unecessary changes to chatcommandpatch.

PEPPERcula:
> i have no idea why it added so many unecessary changes to chatcommandpatch.

Yeah I also had the same issue where I edit like 2 lines of code then it changes the whole file to the point it messes it up.

NikoCat233:
https://github.com/EnhancedNetwork/TownofHost-Enhanced/blob/7f390dace0ed5dedecc32f85ada6f1d025c05924/Patches/GameOptionsMenuPatch.cs#L499

Mind look into this? lol

NikoCat233:
Visual studio is ok btw

NikoCat233 (approved):

PR #1324, opened 2024-12-10
Madmates can get Impostor Add-ons setting

Bri1101:


NikoCat233:
wait why can this pr build

NikoCat233:
To allow madmates to get impostor addon, we need to make sure madmate is assigned before other addons
And I think Impostor only addon is not decided in checkaddonconflicts

NikoCat233:
This gonna be delayed till next canary release

PR #1326, opened 2024-12-11
Make mod fit with dotfuscator

NikoCat233:


NikoCat233:
Major changes that contributor needs to know
1. You now need to assign public override CustomRoles Role => CustomRoles.*** in every role class and addon class
2. You now need to add [Obfuscation(Exclude = true)] for every enum
3. You may need to Clearly include Prefix & Postfix in method name if you are adding a HarmonyPatch

PR #1327, opened 2024-12-14
Add /spectate command

PEPPERcula:


Modules/Extensions/CollectionExtensions.cs
```diff
@@ -14,18 +14,14 @@ public static class CollectionExtensions
...
-    public static TKey GetKeyByValue<TKey, TValue>(this Dictionary<TKey, TValue> dictionary, TValue value)
-    {
-        foreach (KeyValuePair<TKey, TValue> pair in dictionary)
-        {
-            if (pair.Value.Equals(value))
-            {
-                return pair.Key;
-            }
-        }
+    public static Dictionary<TKey, TValue> AddRange<TKey, TValue>(this Dictionary<TKey, TValue> dictionary, Dictionary<TKey, TValue> other, bool overrideExistingKeys = true)
```
NikoCat233, line 17: Why do you need to modify this?

Modules/OptionHolder.cs
```diff
@@ -321,7 +322,7 @@ private enum RatesZeroOne
     public static OptionItem DisableMeeting;
     public static OptionItem DisableSabotage;
     public static OptionItem DisableCloseDoor;
-
+
```
NikoCat233, line 325: Are u using a web editor?
.editorconfig is not applied for you?

Patches/IntroPatch.cs
```diff
@@ -800,11 +800,21 @@ public static void Postfix()
             }
 
             if (PlayerControl.LocalPlayer.Is(CustomRoles.GM)) // Incase user has /up access
+                {
+                    PlayerControl.LocalPlayer.RpcExile();
+                    Main.PlayerStates[PlayerControl.LocalPlayer.PlayerId].SetDead();
+        }
+
+        foreach (byte spectator in ChatCommands.Spectators)
```
NikoCat233, line 808: Assign spectors with GM
At where GM is assigned to host

NikoCat233: If host is not gm, then this code wont run

Patches/ChatCommandPatch.cs
```diff
@@ -1583,6 +1587,37 @@ static Color32 RndCLR()
...
+                case "/spectate":
+                case "/спектейт":
+                case "/观战":
+                case "/espectar":
+                    if (!GameStates.IsLobby)
+                    {
+                        Utils.SendMessage(GetString("Message.OnlyCanUseInLobby"), PlayerControl.LocalPlayer.PlayerId);
+                        break;
+                    }
+
```
NikoCat233, line 1600: Add options:
Max Spectator limit
Host Approval (this may be a little harder. go get a good ide)

NikoCat233 (changes requested):

Modules/Extensions/CollectionExtensions.cs
PEPPERcula: @NikoCat233 the reason why I modified this is because in RoleAssign I added a line of code and i tested it and the game was on loading screen and the game didn’t spawn

NikoCat233:
can you provide a command that make the spectators cancel their previous commands and back to play normal?

PEPPERcula:
I don't see the point of doing that part though bc in my opinion, once they use it, they can immediately use /unspectate for fun which deafeats the purpose of the whole command. Instead they should use this wisely or when they actually need it

PR #1328, opened 2024-12-14
Make updater a async task instead of a coroutine

Pietrodjaowjao:
Coroutines are based on fps, which makes the update slow, this improves by much the update speed by making it completely asynchronous from the game thread and time.


NikoCat233 (approved):

Pietrodjaowjao:
I acidentally had rewritten some parts of the code that could've remained the same as the original, so that messed up the original FIleAccess, but now its fixed

NikoCat233:
Gonna delay this till next canary release

PR #1329, opened 2024-12-16
Roleblocking Changes (Deputy and Baker)

MargaretTheFool:
https://discord.com/channels/1094344790910455908/1311185992534855701
basically make roleblocking go into effect when someone tries to use their ability so it doesn't give away the baker or deputy


NikoCat233:
seems like there are bug reports regarding this in coven build

MargaretTheFool:
> seems like there are bug reports regarding this in coven build

if you were talking about https://discord.com/channels/1094344790910455908/1318669747352502323 that had nothing to do with this

PR #1337, opened 2024-12-29
New RPC Detection: SilentLink [KillNetwork AI]

Luckyheat:
SilentLink [KillNetwork AI] {1480309}

(In case, if the KillNetwork AI hacked/Intercepted for use by unauthorized persons)
Protection against servers intrusion

### DIFF
--- a/Modules/EAC.cs
+++ b/Modules/EAC.cs
@@ -261,7 +261,7 @@ internal class EAC
                         return true;
                     }
                     break;
-                 case 1480309: // KillNetwork AI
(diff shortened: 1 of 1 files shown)


NikoCat233:
what is this lol
btw u need to add unchecked for non byte values i guess

Luckyheat:
> what is this lol btw u need to add unchecked for non byte values i guess

We can use both methods for AI detection 🤦‍♂️

redphantom1000:
interesting

PR #1339, opened 2024-12-30
Fix "Fake Tasks:" missing on task panel

impostor4291:
title says it all


NikoCat233:
can you provide a example of this bug?

impostor4291:
> can you provide a example of this bug?
the first screenshot was taken before the bug was fixed
and the second one was taken after the bug was fixed
beforethebugfixed
afterthebugfixed

PR #1345, opened 2025-01-02
Fix missing comma for MinerInfo in en_US.json

impostor4291:
so that's why the entire language file was messed up...


NikoCat233:
I manually fixed it

PR #1363, opened 2025-01-10
Add New Tag System alongside Old One

MargaretTheFool:
https://discord.com/channels/1094344790910455908/1326991327614861362


MargaretTheFool:
people had requested a variable for users to /dump for the host (for auto hosting purposes) but I couldn't code that because of how the DumpLog method works

PR #1367, opened 2025-01-14
HexMasterInfoLong fix

Limeau:
The character "乂" was causing HexMasterInfoLong to show up as a string


KingPanda360:
Sadly, it will be hard for the new players to tell what the hex master symbol is now. But they still have the list of icons to refer to(if they know the icon list is a thing).

0xDrMoe:
I don’t think generalizing is the best option. Try an escape key: \

NikoCat233:
image
what?

Limeau:
interesting, on my end it shows differently. Ill close it for now

PR #1370, opened 2025-01-17
Fix overseer interactions

impostor4291:
Fixed bug where modded Overseers couldn't reveal the roles of Tricksters and Illusionist's targets
Also fixed font size of role texts when vanilla Overseers revealed the roles of Tricksters and Illusionist's targets


NikoCat233 (approved):

impostor4291:
I spent 1 hour trying to identify what caused modded Overseers to be unable to see random role texts in meetings...

PR #1373, opened 2025-01-17
Tag Overhaul Sorting

NotPyro404:
Margs Tag Overhaul is epic but can get messy, so this PR should make it so friendCode#1234.txt file can be read in a folder  (Example like Friends folder, Staff folder, ABCDEF folder, etc), as long as its under the Tags directory. The only folders it shouldn't read are MOD_TAGS, VIP_TAGS, & SPONSOR_TAGS.

### DIFF
--- a/Modules/TagManager.cs
+++ b/Modules/TagManager.cs
@@ -23,8 +23,8 @@ namespace TOHE
                 if (!Directory.Exists(@"TOHE-DATA/Tags"))
                     Directory.CreateDirectory(@"TOHE-DATA/Tags");
 
-                // Ensure subfolders (1-5) exist for permission levels
-                for (int i = 1; i <= 5; i++)
+                // Ensure subfolders (0-5) exist for permission levels
+                for (int i = 0; i <= 5; i++) // Changed to 0 through 5
                 {
                     string folderPath = @$"TOHE-DATA/Tags/{i}";
                     if (!Directory.Exists(folderPath))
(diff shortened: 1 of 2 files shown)


MargaretTheFool:
build failed :ten_uhh:
also its permissions 0-5 not 1-5

NotPyro404:
> build failed :ten_uhh: also its permissions 0-5 not 1-5

oh also idk how to build anything im just an idiot who likes doing things sometimes and then isnt surprised they fail

NotPyro404:
I did it I'm a genius

Pietrodjaowjao:
Why is this so hardcoded

MargaretTheFool:
that's what  I was trying to prevent

NotPyro404:
> Why is this so hardcoded

how so? is it not the same system but sorted into specific folders?

the current ones good but can get flooded

NotPyro404:
I am failing to see how this is hardcoded if you could just move the friendcode.txt to a seperate folder to change their permission level, the variables/access lines inside the txt file should stay the same unless i messed that up

NikoCat233 (approved):

NotPyro404:
Okay so now it should use the new system (margs new one, not the one utilizing folders for perms), and read the `friendcode#1234.txt` as long as its under the `Tags directory`, even if it's in a folder (Example like Friends folder, Staff folder, ABCDEF folder, etc). The only folders it shouldn't read are `MOD_TAGS`, `VIP_TAGS`, & `SPONSOR_TAGS`.

PR #1374, opened 2025-01-17
Remove Incompatible Add-Ons when Changing Roles in most situations

MargaretTheFool:
for example if Antidote can only be assigned to neutrals and amnesiac remembers a coven role then antidote is removed from them


NikoCat233:
why not just include this function in set custom role btw

MargaretTheFool:
why didn't I think of that 💀

Pietrodjaowjao:
> why not just include this function in set custom role btw

addition farm lmao

NikoCat233 (approved):

PR #1378, opened 2025-01-21
Fixed some bugs related to RpcSetCustomRole

impostor4291:
Bugs fixed(probably,needs testing):
Recruiting roles/Monarch could not recruit/knight players
Workhorse and Last Impostor could not be assigned
Copycat couldn't copy team-changing add-ons despite the setting being on


NikoCat233:
I will look into this later

NikoCat233:
I think addons should be handled seperatedly from main roles, they should both be handled in set custom role
Like in setcustomrole we check whether the addon is a convert addon
if it is we call switch addons and then check addon conflicts
if not we force set it and then check conflicts

NikoCat233:
add bool force_set_addon = true, bool shouldcheckconflicts = true for setcustomrole to use

impostor4291:
> I think addons should be handled seperatedly from main roles, they should both be handled in set custom role Like in setcustomrole we check whether the addon is a convert addon if it is we call switch addons and then check addon conflicts if not we force set it and then check conflicts

i think i did it in another way?
or I still have to add forceSetAddon?

screenshot 2

impostor4291:
this is the 1st screenshot
screenshot 1

NikoCat233:
image
Then why do you need to manually add false here

impostor4291:
> image Then why do you need to manually add false here

I didn't realize that until I finished these

> this is the 1st screenshot screenshot 1



> i think i did it in another way? or I still have to add forceSetAddon? screenshot 2

if it's ok I'm gonna do it in a new PR

PR #1379, opened 2025-01-21
Fix Vigilante not being able to kill Crewmates to become a Madmate

Bri1101:
This was also related to the RpcSetCustomRole, where it was preventing the Vigilante from killing Crewmates.


NikoCat233:
I will look into this later

NikoCat233:
Check #1378 
gonna delay this

PR #1389, opened 2025-01-23
Make Rift Portal and Firework visible for dead players

Tommy-XL:
Since the players are dead anyway, it would be better to show them the location of the abilities of some roles


Tommy-XL:
Bruh, code alredy do this lol

PR #1391, opened 2025-01-23
Allow getting role from multi languages with chat commands

NikoCat233:
Make /role /bt & other commands not only recognize role names in Host's language


NikoCat233:
whats wrong with this shit? I simply squashed commits from dev2.2.0

PR #1393, opened 2025-01-24
Fix another bug related to RemoveIncompatibleAddons

impostor4291:
an example of this bug:
If an impostor with Last Impostor changed their role to another Impostor role mid-game,the Last Impostor would be removed

(I wanna swear)


impostor4291:
why do i feel the previous fix is pointless

NikoCat233:
I think i fixed this

PR #1394, opened 2025-01-24
Guesser

impostor4291:


impostor4291:
forgot to list the changes...
Impostor/Madmate guessers can no longer guess Madmate roles if options are off
Fix Nice Guesser cannot guess any Crewmate players when Nice Guesser's "Can Guess Crewmate Roles" is off

NikoCat233:
will player.isplayerteam*** work in these codes

PR #1395, opened 2025-01-24
New Option For Low Load: Delay Update Names

Tommy-XL:
New low load option

The main purpose of this setting is to update player names on the host only when NotifyRoles was runned for vanilla players, so it is not necessary for it to update the nickname every second

Works only during the game

For modded clients that are not hosts, this is not necessary, because the main load goes only to the host

### DIFF
--- a/Modules/LocateArrow.cs
+++ b/Modules/LocateArrow.cs
@@ -171,7 +171,7 @@ static class LocateArrow
             }
         }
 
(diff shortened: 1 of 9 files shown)


Tommy-XL:
Haven't tested it yet and set the setting to off by default

Tommy-XL:
Okay, after testing it works as it should
The only question is, should this setting be enabled by default?

NikoCat233 (approved):

PR #1397, opened 2025-01-24
Improve AntiBlackout

Tommy-XL:
Players who have the kill button should no longer change role for self, unless they are dead

This eliminates the need to reset kill cooldown using `player.RpcGuardAndKill()`
Instead, the AU code itself does it


Tommy-XL:
Seems works fine

Tommy-XL:
Same after release next beta?

NikoCat233:
Nah im just doing code review

NikoCat233:
are u sure sending game data after desync setting roles is fine?
shouldnt we be sending spoof alive game data first, then we set desync roles?

Tommy-XL:
> are u sure sending game data after desync setting roles is fine? shouldnt we be sending spoof alive game data first, then we set desync roles?

This will not work if IsDead is true and try to assign a Crewmate for example, then it will remain dead

Tommy-XL:
According to the conditions of revival, the role must be count as alive and the player must be dead

image

NikoCat233:
`    public static void SetIsDead(bool doSend = true, [CallerMemberName] string callerMethodName = "")
    {
        SkipTasks = true;
        RevivePlayersAndSetDummyImp();
        logger.Info($"SetIsDead is called from {callerMethodName}");
        if (IsCached)
        {
            return;
        }
        isDeadCache.Clear();
        foreach (var info in GameData.Instance.AllPlayers)
        {
            if (info == null) continue;
            isDeadCache[info.PlayerId] = (info.IsDead, info.Disconnected);
            info.IsDead = false;
            info.Disconnected = false;
        }
        IsCached = true;
        if (doSend) SendGameData();
    }`
    
    Wont send game data send host's view to other players?
    Does this conflicts with previous RevivePlayersAndSetDummyImp

Tommy-XL:
Well, I don’t quite understand the question, but the main thing is that the code works as it should
I tested it several times with different conditions

NikoCat233:
ok then

PR #1413, opened 2025-01-29
No check limit add-ons in RemoveIncompatibleAddOns

Tommy-XL:
I feel like the `RemoveIncompatibleAddOns` code need recode

### DIFF
--- a/Modules/ExtendedPlayerControl.cs
+++ b/Modules/ExtendedPlayerControl.cs
@@ -64,7 +64,7 @@ static class ExtendedPlayerControl
 
         foreach (var addon in roles)
         {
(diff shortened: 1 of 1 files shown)


NikoCat233:
I do agree with this. But current codes do work

Tommy-XL:
For now, I imagine the following check structure: First, the code checks the condition whether it is necessary to check for a conflict, and if it is necessary, then the new add-on is checked with the old ones, and if the old add-on is not compatible, then it needs to be removed, after which the new add-on is assigned via `SetSubRole()`

NikoCat233 (approved):

PR #1425, opened 2025-01-31
Shapeshift and Flash

Reborn5537:
CustomRolesHelper Change
Game experience modification about morphing class

### DIFF
--- a/Modules/CustomRolesHelper.cs
+++ b/Modules/CustomRolesHelper.cs
@@ -1088,7 +1088,10 @@ public static class CustomRolesHelper
                     || pc.Is(CustomRoles.Spurt)
                     || pc.Is(CustomRoles.Chameleon)
                     || pc.Is(CustomRoles.Alchemist)
-                    || pc.Is(CustomRoles.Mare))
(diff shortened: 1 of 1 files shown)


MoeSmellsHA:
EW! YUCK

MoeMONKEY936:
EW YUCK

NikoCat233:
why do you have to prevent shapeshifter from getting flash

Reborn5537:
Maybe it's because I'm lazy or my brain doesn't work very well when I'm writing this stuff lol
But one thing is for sure is that professions like shapeshifter class getting flash will directly reveal their identity (which makes no sense)
(Maybe we can adjust the speed during morphing to the same model of the morphing object? I'm thinking about writing it that way, so I'll leave it that way for now)

Reborn5537:
66666666你去玩一局试试

Reborn5537:
千面鬼和变形者有闪电侠等于白板

Reborn5537:
发挥个shit功能(正义赌怪,末日赌怪: :D)

PR #1431, opened 2025-02-01
Madmate Roles can be Last Impostor

impostor4291:
https://discord.com/channels/1094344790910455908/1335076519898255411


NikoCat233:
why not just use IsPlayerImpostorTeam and check whether player has a kill button to use

NikoCat233 (approved):

PR #1432, opened 2025-02-01
Add Enchanted in more roles

MargaretTheFool:
https://discord.com/channels/1094344790910455908/1335350695624314971
also changed the neutral color in Necroview and Visionary so people don't have to do [this](https://cdn.discordapp.com/attachments/1094351226486915096/1330672526975696946/Screenshot_678.png?ex=679f4f9d&is=679dfe1d&hm=1308200a69ddc892b2031c18bf7c8a6b575876038bf147ecbec90f0ea328187b&)

### DIFF
--- a/Roles/AddOns/Common/Necroview.cs
+++ b/Roles/AddOns/Common/Necroview.cs
@@ -31,7 +31,7 @@ public class Necroview : IAddon
                 or CustomRoles.Egoist
                 or CustomRoles.Recruit
                 or CustomRoles.Soulless)
-                return Main.roleColors[CustomRoles.Knight];
+                return "7f8c8d";
         }
 
         if ((customRole.IsImpostorTeamV2() || customRole.IsMadmate() || target.Is(CustomRoles.Madmate)) && !target.Is(CustomRoles.Admired))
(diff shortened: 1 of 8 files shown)


Roles/AddOns/Common/Necroview.cs
```diff
@@ -31,7 +31,7 @@ or CustomRoles.Contagious
                 or CustomRoles.Egoist
                 or CustomRoles.Recruit
                 or CustomRoles.Soulless)
-                return Main.roleColors[CustomRoles.Knight];
+                return "7f8c8d";
```
NikoCat233, line 34: why do you use static role color here?
maybe some people will use custom role color?

NikoCat233 (changes requested):

Roles/AddOns/Common/Necroview.cs
impostor4291: if you have custom role color file for your lobby,you will know why
(I have one for my lobby,and Knight's color is changed to #964b00(brown).
Neutrals appear with brown colored names instead of gray colored names)

MargaretTheFool: that's exactly why

NikoCat233:
weired. The conflicts disappeared itself

PR #1435, opened 2025-02-02
Medic: change role type to Crewmate after set protect & Some changes in code

Tommy-XL:


Tommy-XL:
Merge this after the release of 2.2.0

PR #1437, opened 2025-02-03
Change Notify Roles

Tommy-XL:
Prevent case of anti-cheat kick


NikoCat233:
Provide a option that disable this slow notify roles
So we bear trash anticheat on Sloths officals, and play freely on modded servers

Tommy-XL:
> Provide a option that disable this slow notify roles So we bear trash anticheat on Sloths officals, and play freely on modded servers

I made the code so that it does not load the host, because there are extra `NotifyRoles` in the code

Tommy-XL:
Wtf

image

Tommy-XL:
How it's happened

Tommy-XL:
Open new

PR #1438, opened 2025-02-03
CheckGameEndPatch

impostor4291:
Fix bugs related to additional winners(don't ask me why i added a for loop and a void)
[https://discord.com/channels/1094344790910455908/1333490802407768137
https://discord.com/channels/1094344790910455908/1259921051962572840(oh nvm)]
Madmate roles can no longer win with Neutrals if "All neutrals win together" is on
Jackal can no longer win by sabotage if a Madmate role is alive(or if Madmate's count mode is set to Impostors and a player with Madmate addon is alive)


Patches/CheckGameEndPatch.cs
```diff
@@ -470,10 +359,138 @@ public static bool Prefix()
                         Logger.Info($"Removed {pc.GetNameWithRole()} from winner ids", "Hurried Win Check");
                     }
                 }
-            }
+                
+                int PossibleAdtWinners = Main.AllPlayerControls.Count(x => x.GetCustomRole().IsNB() || x.Is(CustomRoles.Lovers));
```
NikoCat233, line 363: I still need to ask, wtf is this

NikoCat233 (changes requested):

impostor4291:
> why do you need a loop? The void seems only needed to be run once?

this is why loop is needed <:ten_omg_cry:1156283894576975952>
Romantic followed Opportunist.Opportunist followed Crewmates victory
Romantic should have followed Opportunist victory,but...
image

Patches/CheckGameEndPatch.cs
impostor4291: > I still need to ask, wtf is this

Idk how many times...nvm I'm gonna let Main.AllPlayerControls.Length decide

NikoCat233 (approved):

NikoCat233:
Though the code looks bad, it works at least.

PR #1444, opened 2025-02-05
Global Bypass RateLimit

NikoCat233:
1. Cook a buffer like fixed update buffer for Rpc teleport
This makes sure that only 5 snapto is sent with reliable per second for a single player, while the other is sent with none
2. Send RoleSenders with a 0.3f delay per player on officials.
We need to keep set role in reliable. I guess this still cause kicks
3. Force send all Custom Rpcs with None by patching StartRpcImmediately
4. /xf set name locally for host. Use rpc set name private for clients.
5. rpc set name private now send with none on officials

### DIFF
--- a/Patches/ChatCommandPatch.cs
+++ b/Patches/ChatCommandPatch.cs
@@ -1032,26 +1032,26 @@ internal class ChatCommands
                     canceled = true;
                     Utils.SendMessage(GetString("Message.CanNotUseByHost"), PlayerControl.LocalPlayer.PlayerId);
                     break;
-
-                case "/xf":
-                case "/修复":
-                case "/修":
-                    canceled = true;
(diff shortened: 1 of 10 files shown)


Patches/ChatCommandPatch.cs
```diff
@@ -1044,8 +1044,7 @@ public static bool Prefix(ChatController __instance)
                     }
                     foreach (var pc in Main.AllPlayerControls)
                     {
-                        if (pc.IsAlive()) continue;
-
+                        if (!pc.IsAlive()) continue;
```
Tommy-XL: No, need skip alive here

Tommy-XL (approved):

Tommy-XL (approved):

PR #1446, opened 2025-02-05
Check last closest vent

Tommy-XL:
Check only player who have blocked vent
If player again have access for vent `PlayerHadBlockedVentLastTime` check this

### DIFF
--- a/Patches/PlayerControlPatch.cs
+++ b/Patches/PlayerControlPatch.cs
@@ -1071,7 +1071,7 @@ class FixedUpdateInNormalGamePatch
 {
     private static readonly StringBuilder Mark = new(20);
     private static readonly StringBuilder Suffix = new(120);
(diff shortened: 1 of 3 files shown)


NikoCat233:
The original already serialize with SendOption.None and is only updated when a player has walked to a new vent

NikoCat233:
Oh sorry im dumb. Nevermind.

NikoCat233:
Use buffertime in player control fixed update
delay it with 5 ticks

NikoCat233:
This doesnt really matter.
Since no matter what, the system is still updated 30 times per second

Tommy-XL:
> Use buffertime in player control fixed update delay it with 5 ticks

Like `if (buffertime == 10)` ?

NikoCat233:
buffertime % 6 == 0

PR #1455, opened 2025-02-07
Android support

Tommy-XL:
Since I can build AmongUs.GameLibs.Android I can fix the current problems


Tommy-XL:
Since at the moment we cannot check the Android version, I simply made corrections that prevent the assembly from being assembled for Android

And since we are testing the beta version, soon perhaps one of us will continue correcting the code to support Android

Tommy-XL:
Since at the moment we cannot check the Android version, I simply made corrections that prevent build for Android

And since beta testing is coming soon, perhaps one of us will continue to fix the code to support Android.

Pietrodjaowjao:
Needs changes related to creating TOHE-Data folder, obviousily the folder get on pc version is inacessible in android

Tommy-XL:
I will open new later

PR #1470, opened 2025-02-12
Doomsayer Rework

MargaretTheFool:
(Defaults to off) 
Doomsayer can now observe a player to get a hint about their role upon the next meeting


Tommy-XL:
In fact, I think that this is a very inefficient code that has 2 significant disadvantages

When adding a new role, probably no one will add the new role to the Doomsayer list

Even if someone adds a role to the list, they will also have to change the lines and this will reset the current lines in all languages, which is very annoying to change this every time

Tommy-XL:
So, I believe that the code should show the code the same as in TOU-R when it selects a player and from the randomly active roles only one is correct

Roles/Neutral/Doomsayer.cs
```diff
@@ -268,4 +287,543 @@ public void SendMessageAboutGuess(PlayerControl guesser, PlayerControl playerMis
             }, 0.7f, "Doomsayer Guess Msg 2");
         }
     }
+    public override bool CanUseKillButton(PlayerControl pc) => EasyMode.GetBool();
+    public override void SetKillCooldown(byte id) => ObserveCooldown.GetFloat();
```
Tommy-XL, line 291: Use
`Main.AllPlayerKillCooldown[id] = ObserveCooldown.GetFloat();`

MargaretTheFool:
> So, I believe that the code should show the roles the same as in TOU-R when it selects a player and from the randomly active roles only one is correct
> 
> Or does it work somehow differently and I don’t remember how Doomsayer in TOU-R works

TOU-R does the same thing I've done, roles are put into categories for doomsayer and are hard-coded, which is fine because that mod doesn't have as much roles
maybe I could add a DoomsayerCategory enum in RoleBase and just assign it in every role?

MargaretTheFool:
I can perfectly understand why fortune Teller was reworked to the way it is now but if I did the same thing with doomsayer that would make the role way too easy

Tommy-XL:
It might be better to make the code check each role via `Custom_RoleType` to avoid manual addition, and also to avoid adding each role to the lines it would be better to make the code do this, for example, as this code does in `FortuneTeller.GetTargetRoleList`

MargaretTheFool:
the issue with that is that im not sorting it by `Custom_RoleType`
if you think I should then I will, but I don't really want to

PR #1471, opened 2025-02-13
Fix TNA's dying to roles they shouldnt

MargaretTheFool:
https://discord.com/channels/1094344790910455908/1318662488962629632
in a very niche situation, if Penguin is on a ladder while dragging a TNA, the TNA will die because I don't know how the code works for that
everything else should be fixed
Hex Master, AbyssBringer, Kamikaze, and Witch all have the option to kill TNA's every other role does not (defaults to off)
Pirate will still get its win counted but the target wont die if TNA
Bodyguard will ignore the kill
Jailer's execution wont actually kill if TNA


Roles/Impostor/Penguin.cs
```diff
@@ -207,10 +213,11 @@ public override void OnFixedUpdate(PlayerControl penguin, bool lowLoad, long now
...
             if (AbductTimer <= 0f && !penguin.MyPhysics.Animations.IsPlayingAnyLadderAnimation())
             {
-                // Set IsDead to true first (prevents ladder chase)
-                AbductVictim.Data.IsDead = true;
-                AbductVictim.Data.MarkDirty();
-
+                if (!AbductVictim.IsTransformedNeutralApocalypse()) { 
+                    // Set IsDead to true first (prevents ladder chase)
+                    AbductVictim.Data.IsDead = true;
+                    AbductVictim.Data.MarkDirty();
```
Tommy-XL, line 219: This won't work, it's better to add a check to `OnCheckMurderAsKiller`

Roles/Impostor/Penguin.cs
```diff
@@ -140,6 +140,12 @@ public override void OnReportDeadBody(PlayerControl reporter, NetworkedPlayerInf
     {
         stopCount = true;
         // If you meet a meeting with time running out, kill it even if you're on a ladder.
+        if (AbductVictim.IsTransformedNeutralApocalypse())
+        {
+            RemoveVictim();
+            Logger.Info($"{AbductVictim.GetRealName()} is TNA, no meeting kill", "Penguin");
+            return;
+        }
```
Tommy-XL, line 148: When AbductVictim is null it will cause an error

Roles/Impostor/Penguin.cs
MargaretTheFool: are you suggesting that penguin just cant drag tna's?

Roles/Neutral/Shocker.cs
```diff
@@ -189,7 +189,7 @@ private void OnFixedUpdateShocker(PlayerControl player, bool lowLoad, long nowTi
         {
             foreach (Collider2D collider in shockedRooms)
             {
-                if (collider.IsTouching(player.Collider))
+                if (collider.IsTouching(player.Collider) && !player.IsTransformedNeutralApocalypse())
```
Tommy-XL, line 192: Why not do the check outside of foreach

Roles/Neutral/Revolutionist.cs
```diff
@@ -234,7 +234,7 @@ private static void OnFixUpdateOthers(PlayerControl player, bool lowLoad, long n
                     SetDrawPlayerRPC(player, rv_target, true);
                     NotifyRoles(SpecifySeer: player, SpecifyTarget: rv_target);
                     ResetCurrentDrawTarget(playerId);
-                    if (IRandom.Instance.Next(1, 100) <= RevolutionistKillProbability.GetInt())
+                    if (IRandom.Instance.Next(1, 100) <= RevolutionistKillProbability.GetInt() && rv_target.IsTransformedNeutralApocalypse())
```
Tommy-XL, line 237: here should be `!` ?

Tommy-XL:
Well probably everything seems fine, lol

PR #1472, opened 2025-02-13
Make Glitch As Desync Shapeshifter

Tommy-XL:


Tommy-XL:
Lol I'm convinced once again that the current UnShift code is so broken that the player who have UnShift button does not return the glitch skin when he use Shapeshift

Tommy-XL:
And with this #1463 code everything works fine

PR #1474, opened 2025-02-13
Move CheckAdditionalWinners

Tommy-XL:
Will each iteration of the loop run the same code?
For what?


NikoCat233:
Impy want to check additional winners multiple times
so if an additional is based on another additonal winner, it wont miss

PR #1451, opened 2025-02-06
Exile screen

impostor4291:
New option:Show betrayal addons of ejected players(https://discord.com/channels/1094344790910455908/1295165639530643486)

Count Madmate roles towards "X Impostors remain(s)" on ejection
(For example,if an Impostor and a Parasite are alive,it will show "2 Impostors remain")
Add a reminder of there being only converted crewmates left when they keep the game going
also added new strings for showing remaining NAs and Covens

PR #1476, opened 2025-02-13
Add ConfirmEjectionPatch

Tommy-XL:
The current code changes the player's name via LateTask, I think it would be better if the code was run even when the meeting forced ended

And need wait merged  #1451


Tommy-XL:
Time do this

PR #1484, opened 2025-02-17
add addon changes in rolelog

NikoCat233:


NikoCat233:
Needs recode. 
Should directly record display name, associated with addon changes

PR #1487, opened 2025-02-20
Show Available Update

Tommy-XL:
In launching the game, a notify window pops up when an update is available.
Appears only once per launch and only in Beta and full releases builds

image

### DIFF
--- a/Modules/ModUpdater.cs
+++ b/Modules/ModUpdater.cs
@@ -438,11 +438,14 @@ public class ModUpdater
     {
         if (InfoPopupV2 != null)
         {
-            InfoPopupV2.Show(message);
-            InfoPopupV2.transform.FindChild("Background").localScale *= 2f;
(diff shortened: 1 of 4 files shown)


NikoCat233:
mod updater check release time on github and compare it with local version number.
the problem is that a release is always published 1-3 days later after it is built.
Can u help add a static file in main branch, let mod updater read the static file to check for updates.
Also we can define download url in it and ise multiple endpoints for update

Tommy-XL:
> mod updater check release time on github and compare it with local version number. the problem is that a release is always published 1-3 days later after it is built. Can u help add a static file in main branch, let mod updater read the static file to check for updates. Also we can define download url in it and ise multiple endpoints for update

I don't quite know how Mod Manager works, so I don’t quite understand what it’s for

NikoCat233:
ok j will do it myself later

NikoCat233 (approved):

NikoCat233 (approved):

PR #1491, opened 2025-02-21
Eraser Rework

Bri1101:
Eraser now selects a player with their kill button instead of voting.
https://discord.com/channels/1094344790910455908/1336444758679359518


NikoCat233:
Addon assign is changed. Will need to adapt to it.

PR #1517, opened 2025-03-04
Coven Improvements

MargaretTheFool:
Conjurer gets two new settings: Kill Cooldown and Reset Targets After Meetings

Also now all ghosts can see coven icons


Patches/MeetingHudPatch.cs
```diff
@@ -736,8 +736,10 @@ public static bool Prefix(MeetingHud __instance, byte srcPlayerId, byte suspectP
         }
 
         // Coven Necronomicon Voting
-
-        if (suspectPlayerId == 253 && voter.IsPlayerCoven())
+        if (voter.IsPlayerCoven() && Main.AllAlivePlayerControls.Where(x => x.IsPlayerCoven()).Count() == 1) {
```
Tommy-XL, line 739: `Where` and `Count` can be combined because `Count`, `Any`, ect. can perform the `Where` function
So it's better to use
```suggestion
        if (voter.IsPlayerCoven() && Main.AllAlivePlayerControls.Count(x => x.IsPlayerCoven()) == 1) {
```

NikoCat233:
Conflicts

PR #1526, opened 2025-03-12
Requiter(Knight hidden role)

impostor4291:
https://discord.com/channels/1094344790910455908/1319776581371297792


Roles/Crewmate/Knight.cs
```diff
@@ -48,3 +60,58 @@ public override bool OnCheckMurderAsKiller(PlayerControl killer, PlayerControl b
...
+        if (exiled == null || exiled.Object == null || exiled.Object == player || !player.IsAlive()) return;
+        if (exiled.Object.IsPlayerCrewmateTeam())
+            player.RpcIncreaseAbilityUseLimitBy(1);
+    }
+
+    public override bool ForcedCheckMurderAsKiller(PlayerControl killer, PlayerControl target)
+    {
+        if (target.Is(CustomRoles.Solsticer)) return true;
+        if ((target.Is(CustomRoles.NiceMini) || target.Is(CustomRoles.EvilMini)) && Mini.Age < 18) return true;
+        if (Knight.RequiterCanKillTNA.GetBool())
```
NikoCat233, line 102: what does this check?

NikoCat233 (changes requested):

Roles/Crewmate/Knight.cs
impostor4291: it checks the setting
if it's enabled,requiter ignores all protections(except mini and solsticer)...wait i should probably rename the setting

PR #1540, opened 2025-03-25
Patch RpcClose to send setname calls

NikoCat233:
Doesnt work

Some function can call MeetingHud.Close without using RpcClose.

I can't find this function in dnSpy


NikoCat233:
Done in dev 2.3.0 by patching meetinghud.update

PR #1559, opened 2025-04-18
Rework role map Place holder

NikoCat233:
Due to the fact that antiblackout no longer works on role replacer, and current mod codes mess up with role replacer, we should be working on removing role replacer.

For a crew basis, we can serialize everyone to them as a crew, no need to serialize real impostors to them since antiblackout no longer works with real impostors
For a impostor, we serialize real impostor to them
For a desync killer, we serialize everyone as crew

The delayed update of role map cause engineer role basis to lose their vent button, which should be fixed

We will require every player to have a role class (rpc set custom role) before setting vanilla roles, in this way we can get roles realtime and 100% send them correct roles.


NikoCat233:
I give up completely changing role map stuffs after looking into the codes. It is too complicated.
So i choose to reset rolemap after every antiblackout, and the new role map is based on real time RoleClass instead of static role class. In this way we can simply change role mid game with this role base
To use RpcChangeRoleBasis, you should make a flag in Role to change ThisRoleType mid game, and then use RpcChangeRoleBasis to set that role type

PR #1572, opened 2025-05-11
Narc shares guess access with Crewmates instead of Impostors

impostor4291:


Modules/GuessManager.cs
```diff
@@ -119,17 +119,17 @@ public static bool GuesserMsg(PlayerControl pc, string msg, bool isUI = false)
             pc.ShowInfoMessage(isUI, GetString("GuessDead"));
             return true;
         }
-        if (!pc.Is(CustomRoles.NiceGuesser))
+        if (!pc.Is(CustomRoles.NiceGuesser) && !(pc.Is(CustomRoles.EvilGuesser) && pc.Is(CustomRoles.Narc)))
```
MargaretTheFool, line 122: if someone is evil guessser and not narc (which would also make them not be nice guesser), this would trigger because
`!(F) && !(T && F)`
`T && !(F)`
`T && T`
`T`

Modules/GuessManager.cs
```diff
@@ -243,7 +243,7 @@ public static bool GuesserMsg(PlayerControl pc, string msg, bool isUI = false)
 
                 if (!role.IsEnable() && !role.RoleExist(true) && Options.CanOnlyGuessEnabled.GetBool())
                 {
-                    pc.ShowInfoMessage(isUI, string.Format(GetString("GuessRoleNotEnabled"), role.ToString()));
+                    pc.ShowInfoMessage(isUI, string.Format(GetString("GuessRoleNotEnabled"), GetString(role.ToString())));
```
MargaretTheFool, line 246: why is this needed?

Modules/GuessManager.cs
impostor4291: umm

NikoCat233 (changes requested):
Add a option to choose to share guesses with imp or crews

PR #1573, opened 2025-05-11
Add addon conflicts for Narc

impostor4291:
Narc can't be Mare and Mimic
Narc can be Torch
Narc doesn't take an addon spot(um)


MargaretTheFool:
why should narc not take an addon spot when lovers and madmate does?

impostor4291:
> why should narc not take an addon spot when lovers and madmate does?

in some cases imps can tell who narc is by checking how many addons other imps have
for example when all the other imps have 3 addons(max addons is set to 3) while one has only 2 in imps pov
if one imp notices it ,they will easily figure out that the one with 2 addons is narc

PR #1580, opened 2025-05-22
Add a few more sounds for some roles

PEPPERcula:


Roles/Impostor/Exorcist.cs
```diff
@@ -128,6 +129,10 @@ public static void ExorcisePlayer(PlayerControl player)
         if (ExorcismEndOnKill.GetBool() && IsExorcismActive)
         {
             IsExorcismActive = false;
+            foreach (var pc in Main.AllPlayerControls)
+            {
+                RPC.PlaySoundRPC(pc.PlayerId, Sounds.TaskComplete);
```
NikoCat233, line 134: Mind check whether RpcPlayerSoundForAll works?

NikoCat233 (changes requested):

Roles/Impostor/Exorcist.cs
PEPPERcula: From what i tested, if i use "foreach (var pc in Main.AllPlayerControls)" it will just play for everyone which it did or else it will just play for the Exorcist.

PR #1582, opened 2025-05-23
Support 2025.4.15

NikoCat233:
InnerSloth released their official patches for oversized packets and ratelimit stuffs.
However, due to the limit of Il2cpp injection, IGameDataMessage is defined as a object instead of an interface in mod codes.
This means you can't directly create and use your own gamedatamessages with innersloth api.

This pr tries to introduce BaseModdedRpc class in order to make modded custom rpc and data to be converted and sent by innersloth official methods. However upon personal testing, injecteing this method has a very high chance to make the game crash on load (It works fine if it injected successfully though).

Looking for help to make the inject more stable and convert other custom rpcs into the new BaseModdedRpc class.


XtraCube:
You can implement interfaces when injecting types into IL2CPP

PR #1592, opened 2025-06-13
Fixed russian translation

yarber8:
I think that god should be devil bc god has an evil aura


NikoCat233:
Our translation is done on Crowdin. I have manually add your changes.
You can check our discord for translation works if you are interested.
Thank you for help anyway.

